PR #42, opened 2020-04-15
Feature/basic api

norling:
Min idé om grundläggande struktur om vi kör flask/peewee för APIet.
Om det ser ok ut så börjar jag kolla på OpenID lite mer, och dan kan kolla på att göra grundstrukturen för en react-sida.


app/test.py
```diff
@@ -0,0 +1,127 @@
...
+    def setUp(self):
+        db.set_database("herdbook", "localhost", "5432", "herdbook", "insecure")
+
+    def test_genebank(self):
+        """
+        Checks that the genebank table exists and that all the fields in the
+        mapping exists in the database.
+        """
+        self.assertTrue(db.Genebank.table_exists())
+        db.Genebank.select().execute(db.DB_PROXY)
```
jonandernovella: Could we make sure that the models are actually bound to the given DB? I.e: `assert Class._meta.database is db.DB_PROXY`

pontus:
Looks fine for the moment as far as I'm concerned.

In the future (and could be done now) I think we should move to generate the models with pwiz.

norling:
pontus - are you suggesting that the model should be generated on the fly from the database? or just that it should be generated automatically instead of manually?

pontus:
Not on the fly, just automatic.

andkaha:
Hold on, why do you remove the schema file?! How do I create a database to load the data from CSV now? Or did you also write a loading script?

norling:
With python, since we weren't supposed to have both SQL and python.

andkaha:
Hm. What makes it impossible to have both SQL and Python?

norling:
Nothing. But this PR had a change request where the database shouldn't be defined in both python and SQL... The first idea was to auto-generate the python-mapping from the SQL, but that didn't generate the python docstrings, so then I went with generating the SQL from python.

app/herdbook.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
+"""
+This file contains the main webserver of 'the herdbook'.
+
+The server uses Flask to serve a React frontend, connect to a postgres database,
+and do calculations with R.
```
jhagberg: We are not sure about R yet

jhagberg: Not a big issue

app/herdbook.py
```diff
@@ -0,0 +1,87 @@
...
+            },
+        'father': {
+            'id': individual.father.id,
+            'name': individual.father.name
+            },
+        'colour': individual.colour.name,
+        'colour_note': individual.colour_note,
+        'death_date': individual.death_date,
+        'death_note': individual.death_note,
+        'weight_young': individual.weight_young,
```
jhagberg, line 60: Varför heter det weight_young?

app/herdbook.py
norling: true, I can update the comments!

app/herdbook.py
jhagberg: This data is also loaded in weight table according to @kusalananda

app/test.py
jhagberg: Is this relevant still?

PR #50, opened 2020-04-29
Feature/schema tweaks

### LINKED ISSUE #1: Finish schema
Finish the database schema.

andkaha:
Changes:

* Uniqueness constraints on `genebank.name` and `colour.name`.
* In individual, change the unique constraint on `name` to _include_ `genebank_id`.

### DIFF
--- a/sql/schema.sql
+++ b/sql/schema.sql
@@ -73,7 +73,7 @@ CREATE TABLE individual (
 	--        Github issue #12.
 	-- UNIQUE (certificate),
 
-	UNIQUE (number),
+	UNIQUE (number, genebank_id),
 	FOREIGN KEY (genebank_id)	REFERENCES genebank(genebank_id),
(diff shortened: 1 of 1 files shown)


norling (changes requested):
Wouldn't it make more sense to set `individual.certificate` as `UNIQUE` right away, and just not be able to load all the data until it's fixed? To me, making schema changes later seems a lot more involved than reloading the data later.

pontus:
Is the number collision free between the two genebanks? The weakened constraint looks sensible to me (in general).

pontus (approved):
Looks ok.

andkaha:
@norling I think it's unreasonable to block the entire project until the data is fixed. In the Gottlandskanin data set there are 370 missing or duplicated certificates, and in the Mellerud set there are 72.

PR #58, opened 2020-05-19
Add user context

norling:
Attempt to handle user state semi-globally using a user state context.


danr:
Competing PR here: #59

PR #60, opened 2020-05-20
Feature/login

norling:
Add basic session based login system


app/utils/database.py
```diff
@@ -316,3 +330,44 @@ def verify(try_init=True):
...
+
+def register_user(email, password):
+    """
+    Creates a new user from an e-mail and password, returning the new user
+    object.
+    """
+    print("inserting user %s, %s" % (email, password))
+    user = User(email=email,
+                uuid=uuid.uuid4().hex,
+                password_hash=generate_password_hash(password),
```
danr, line 350: :heavy_check_mark: I checked that werkzeug adds a salt to this: https://werkzeug.palletsprojects.com/en/1.0.x/utils/#werkzeug.security.generate_password_hash

app/herdbook.py
```diff
@@ -7,12 +7,28 @@
...
 
 import sys
+import uuid
 import time
 from flask import Flask, render_template, jsonify
 
 import utils.database as db
 
 APP = Flask(__name__, static_folder="/static")
+APP.secret_key = uuid.uuid4().hex
```
danr, line 22: Is this `secret_key` related to CORS?

app/utils/database.py
```diff
@@ -360,7 +368,7 @@ def authenticate_user(email, password):
         # attacks
         check_password_hash("This-always-fails", password)
     logging.info("Failed login attempt for %s", email)
-    return None
+    return User()
```
danr: Why not return a None here? What does this dummy User contain, every field initialized to a default null value? Isn't that potentially problematic?

app/utils/database.py
```diff
@@ -235,6 +235,14 @@ class User(BaseModel):
     validated = BooleanField(default=False)
     privileges = TextField(null=True)
 
+    def frontend_data(self):
+        """
+        Returns the information that is needed in the frontend.
+        """
+        return {'email': self.email,
+                'validated': self.validated if self.validated else False,
```
danr, line 243: What is `validated`? It does not seem to be able to get this value to be True from the python code.

app/herdbook.py
```diff
@@ -29,7 +34,38 @@ def after_request(response):
...
+def login():
+    """
+    Parses a login form and sets session variables when logged in.
+    If login fails the system will default to an anonymous user.
+    """
 
+    form = request.json
+    # Authenticate the user and return a user object
+    user = db.authenticate_user(form.get('username'), form.get('password'))
+    session['user'] = user.frontend_data()
```
danr: We're not storing a `user` here, we're storing user frontend data, so the field could be named to reflect that like:
```
session['user_frontend_data'] = user.frontend_data()
```
This is a bit confusing in the `get_user()` route above where it looks like you either get a user or user frontend data.

frontend/user_context.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react'
+
+import {post, get} from './communication'
+
+/** The currently logged in user, if any */
+export interface User {
+  email: string | null
+  active: boolean
```
danr: The backend sends `validated` not `active`.

frontend/user_context.tsx
```diff
@@ -0,0 +1,83 @@
...
+}
+
+const null_user: User = {
+  email: null,
+  active: false
+}
+
+/** The currently logged in user, if any, and functionality to log in and log out */
+export interface UserContext {
+  user: User
```
danr: I'm a bit skeptical if you see if you are actually logged in on if the `email` is null or string, why not have `user: User | undefined` here?

register_user.sh
```diff
@@ -0,0 +1,44 @@
...
+
+cd "$( dirname "$0" )" || { echo 'cd error' >&2; exit 1; }
+
+if [ ! -e app/config.ini ]
+then
+    cat <<-'END_ERROR' >&2
+	Couldn't find config file 'app/config.ini'
+	Create this file from 'app/config.ini.default' with the credentials
+	to your database and then run this script again.
+	END_ERROR
```
danr, line 26: These heredocs are a bit annoying since they contain tabs. Why not a multi-line string, like:
```
echo "
        Couldn't find config file 'app/config.ini'
￼	Create this file from 'app/config.ini.default' with the credentials
￼	to your database and then run this script again.
"
```

.docker/herdbook-dev
```diff
@@ -1,9 +1,20 @@
+# React build stage
+FROM node:13 as build-stage
+
+COPY ./frontend /build/
+
+WORKDIR /build
+
+RUN yarn install
+RUN yarn run build
```
danr, line 9: Very cool! :sunglasses:

danr (commented):
Jotted down my thoughts and ideas while reading through this code. 

My experience with login systems is limited and I didn't spend much time on trying to find security vulnerabilities. That would take longer time. I do recommend that someone more experienced with such systems take a look on it.

app/herdbook.py
```diff
@@ -20,6 +20,12 @@
 
 APP = Flask(__name__, static_folder="/static")
 APP.secret_key = uuid.uuid4().hex
+# cookie options at https://flask.palletsprojects.com/en/1.1.x/security/
+APP.config.update(
+    SESSION_COOKIR_SECURE = False if APP.config['ENV'] == 'development' else True,
```
danr: `COOKIR` typo

danr:
Another note: the merge commits from other PRs are included. These could be removed to have a linear history.

app/utils/database.py
norling: yeah - I checked that too! :) very convenient.

app/herdbook.py
norling: it's used for signing the session cookie. You often set a static secret (since it allows the same cookies to be used for multiple servers, etc.), but I figured a random one would to for now... It should be moved to the settings file later though.

app/utils/database.py
norling: The idea is that an anonymous user is still a user, and I figured that it would allow the rest of the code to be cleaner if you could trust that the user was always the same type. Not sure it's a good idea, but that was the reasoning!

app/utils/database.py
norling: yeah, there used to be both "active" and "validated", and I confused them when I removed "active". I'll sort it out!

app/herdbook.py
norling: true. I'll try to make this more explicit!

app/herdbook.py
```diff
@@ -22,9 +22,9 @@
 APP.secret_key = uuid.uuid4().hex
 # cookie options at https://flask.palletsprojects.com/en/1.1.x/security/
 APP.config.update(
-    SESSION_COOKIE_SECURE = False if APP.config['ENV'] == 'development' else True,
-    SESSION_COOKIE_HTTPONLY = True,
-    SESSION_COOKIE_SAMESITE = 'Lax' if APP.config['ENV'] == 'development' else 'Strict',
+    SESSION_COOKIE_SECURE=not APP.config['ENV'] == 'development',
```
danr: Or `APP.config['ENV'] != 'development'`

register_user.sh
```diff
@@ -20,10 +20,10 @@ cd "$( dirname "$0" )" || { echo 'cd error' >&2; exit 1; }
...
 then
     cat <<-'END_ERROR' >&2
-	Couldn't find config file 'app/config.ini'
-	Create this file from 'app/config.ini.default' with the credentials
-	to your database and then run this script again.
-	END_ERROR
+    Couldn't find config file 'app/config.ini'
+    Create this file from 'app/config.ini.default' with the credentials
+    to your database and then run this script again.
+    END_ERROR
```
danr: Oh but this won't work, you need tabs for the `<<-`-heredocs to be indented! https://unix.stackexchange.com/questions/76481/cant-indent-heredoc-to-match-nestings-indent

danr (changes requested):

danr:
Ok! :+1: I have nothing more here!

danr (approved):

PR #62, opened 2020-05-25
Feature/router

norling:
I think that react-router-dom seems to do a decent job, so I'm happy with going forward with that.

The cache solution isn't great, but it should work for now at least...

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -5,5 +5,11 @@ RUN yarn install
 VOLUME /frontend/node_modules
 VOLUME /frontend/.cache
 VOLUME /frontend/.parcel-cache
+
+# make sure that stale caches aren't used
+RUN rm -rf /frontend/node_modules/*
+RUN rm -rf /frontend/.cache/*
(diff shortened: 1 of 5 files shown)


frontend/Dockerfile
```diff
@@ -5,5 +5,11 @@ RUN yarn install
 VOLUME /frontend/node_modules
 VOLUME /frontend/.cache
 VOLUME /frontend/.parcel-cache
+
+# make sure that stale caches aren't used
+RUN rm -rf /frontend/node_modules/*
+RUN rm -rf /frontend/.cache/*
+RUN rm -rf /frontend/.parcel-cache/*
+
```
danr: ~See https://github.com/NBISweden/theherdbook/pull/64~

danr (changes requested):
The routing looks good!

frontend/Dockerfile
danr: See #65 :sunglasses:

danr (approved):

PR #64, opened 2020-05-26
Run docker compose with --renew-anon-modules (-V)

danr:
Apparently docker-compose doesn't treat anonymous VOLUMEs the same way as docker does. The `-V` flag to up makes docker-compose respect the dockerfile semantics. References:
https://github.com/docker/compose/pull/5596
https://github.com/docker/compose/issues/4337


danr:
This was too stupid, closing

PR #68, opened 2020-05-27
Feature/frontend skeleton

norling:
Not the most interesting code, but it should provide the main pages to view and find animals.

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -203,16 +203,16 @@ class Individual(BaseModel):
             if self.father else None
         data['colour'] = self.colour.name if self.colour else None
         data['weights'] = [{'weight':w.weight, 'date':w.weight_date}
-            for w in self.weight_set]
+                           for w in self.weight_set] #pylint: disable=no-member
         data['bodyfat'] = [{'bodyfat':b.bodyfat, 'date':b.bodyfat_date}
-            for b in self.bodyfat_set]
+                           for b in self.bodyfat_set] #pylint: disable=no-member
         data['herd_tracking'] = [
             {
                 'herd_id':h.herd.id,
                 'herd':h.herd.name,
                 'date':h.herd_tracking_date
             }
-            for h in self.herdtracking_set
+            for h in self.herdtracking_set #pylint: disable=no-member
         ]
 
         return data
@@ -441,9 +441,9 @@ def get_genebank(genebank_id, user_uuid=None):
         return None
     try:
         genebank = Genebank.get(genebank_id).as_dict()
-        herd_query = Herd(database=DATABASE).select() \
-                                            .where(Herd.genebank == genebank_id)
-        genebank['herds'] = [h.as_dict() for h in herd_query.execute()]
+        query = Herd(database=DATABASE).select() \
+                                       .where(Herd.genebank == genebank_id)
+        genebank['herds'] = [h.as_dict() for h in query.execute()]
         return genebank
     except DoesNotExist:
         return None
@@ -455,8 +455,8 @@ def get_genebanks(user_uuid=None):
     """
     if user_uuid is None:
         return None
-    genebanks_query = Genebank(database=DATABASE).select()
-    return [g.as_dict() for g in genebanks_query.execute()]
+    query = Genebank(database=DATABASE).select()
+    return [g.as_dict() for g in query.execute()]
 
 def get_herd(herd_id, user_uuid=None):
     """
@@ -468,9 +468,9 @@ def get_herd(herd_id, user_uuid=None):
         return None
     try:
         herd = Herd.get(herd_id).as_dict()
-        individuals_query = Individual(database=DATABASE).select() \
-                                .where(Individual.herd == herd_id)
-        herd['individuals'] = [i.short_info() for i in individuals_query.execute()]
+        query = Individual(database=DATABASE).select() \
+                           .where(Individual.herd == herd_id)
+        herd['individuals'] = [i.short_info() for i in query.execute()]
         return herd
     except DoesNotExist:
         return None
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -87,6 +87,12 @@ class BaseModel(Model):
     This class sets the database to use the postgres database.
     """
 
+    def as_dict(self):
+        """
+        Returns the objects key/value pair as a dictionary.
+        """
+        return self.__dict__['__data__']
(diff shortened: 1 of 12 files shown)


frontend/herd.tsx
```diff
@@ -0,0 +1,38 @@
...
+export function Herd() {
+  let { herdId } = useParams();
+  const [herd, setHerd] = React.useState(undefined)
+
+  React.useLayoutEffect(() => {
+    get(`/api/herd/${herdId}`).then(
+      data => setHerd(data.herd),
+      error => console.error(error)
+    )
+  }, [])
```
danr: This will need `herdId` in the list of dependencies:

```React.useEffect(() => ..., [herdId])```

Also we should use `useEffect` and not `useLayoutEffect`, it was just due to a brainfart I started using the layout version. That one is only useful if you want to synchronously do something with the layouting, such as measuring sizes of dom nodes (like `getClientRect()`)

File: frontend/individual.tsx [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+export function Individual() {
+  let { individualId } = useParams();
+  const [individual, setIndividual] = React.useState(undefined)
+
+  React.useLayoutEffect(() => {
+    get(`/api/individual/${individualId}`).then(
+      data => {console.debug(data); setIndividual(data.individual)},
+      error => console.error(error)
+    )
+  }, [])
```
danr: Same here

frontend/herd.tsx
```diff
@@ -0,0 +1,38 @@
...
+    {herd
+      ? <>
+          <h2>{herd.name ?? `Besättning ${herd.id}`}</h2>
+          <ul>
+            {herd.individuals.map(individual => {
+              return <Link key={individual.id} to={`/individual/${individual.id}`}><li>{individual.name}</li></Link>
+            })}
+          </ul>
+        </>
+      : ''}
```
danr: You can do `{herd && <> ... </>}` instead of `{herd ? <> ... </> : ''}`. False values are already not included in the dom (such as `undefined` here)

frontend/main.tsx
```diff
@@ -42,6 +47,18 @@ function Main() {
...
+              <Genebanks/>
+            </Route>
+            <Route path="/genebank/:genebankId">
+              <Genebank/>
+            </Route>
+            <Route path="/herd/:herdId">
+              <Herd/>
+            </Route>
+            <Route path="/individual/:individualId">
+              <Individual/>
```
danr: I would prefer to forward the routing information to the subcomponents here as props earlier. I don't know if the routing lib  supports this out of the box but one could imagine an API like

```
<Route path="/individual/:id">
  {({id}) => <Individual id={id}/>}
</Route>
```
This way the name `individualId` does need to span several files. This is good because it is not type-checked. That is what I want to achieve.

frontend/herd.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { get } from './communication';
+
+/**
+ * Shows herd information, with a list of all individuals belonging to that
+ * herd.
+ */
+export function Herd() {
+  let { herdId } = useParams();
+  const [herd, setHerd] = React.useState(undefined)
```
danr: This will give a type error later since it will make he type of `herd` be `undefined`. One can get around this by making type interfaces for what it will return:

```
interface Herd {
   ...
}

... const [herd, setHerd] = React.useState(undefined as undefined | Herd)
```

or just saying that it's any if you prefer it to be untyped:

```
... const [herd, setHerd] = React.useState(undefined as any)
```

app/utils/database.py
```diff
@@ -87,6 +87,12 @@ class BaseModel(Model):
     This class sets the database to use the postgres database.
     """
 
+    def as_dict(self):
+        """
+        Returns the objects key/value pair as a dictionary.
+        """
+        return self.__dict__['__data__']
```
danr, line 94: Is this `__data__` set by the SQL lib? Is this the same as `self.__data__`? Is there no library function that can be used instead? `self.get_data()` or something.

app/utils/database.py
```diff
@@ -379,5 +430,60 @@ def fetch_user_info(user_id):
...
+    individuals belonging to that herd. The returned data is limited to the
+    access permission of the user identified by `user_uuid`.
+    """
+    if user_uuid is None:
+        return None
+    try:
+        herd = Herd.get(herd_id).as_dict()
+        individuals_query = Individual(database=DATABASE).select() \
+                                .where(Individual.herd == herd_id)
+        herd['individuals'] = [i.short_info() for i in individuals_query.execute()]
```
danr: I would just call `individuals_query` `query` (or even `q`) since its scope is just two lines :nerd_face: Then it's more apparent that the code is copied-and-pasted if the same name is used in `get_genebanks` above too.

frontend/main.tsx
danr: Another benefit of this which is even better is that the `Individual` component will be easier to test since you won't need to set up routing to test it.

danr (changes requested):
Good stuff! 

I made some general remarks but the main things I would like to see adressed is the part about routing passing information across files in useParams and that there is a type error from the undefined in useState (and this can be fixed correctly here by just adding as any)

A Caveat is that I couldn't test with actual genebank data because I don't have any:P I might find something more when I can actually try it with some data.

app/utils/database.py
norling: I honestly didn't find any. Here're the docs: http://docs.peewee-orm.com/en/latest/peewee/api.html#Model - and as far as I can tell you're "supposed" to handle the result data as objects, but those aren't json serializable...

app/utils/database.py
norling: Haha - it's honestly part of an at times misguided attempt to use more descriptive variable names! 😁 I agree that `query` is a better name though!

frontend/main.tsx
norling: yeah, that would be way neater! I was honestly surprised that I had to maintain variable name as well, but I guess it's a react thing or something... I haven't found any hacky syntax for this though...

frontend/herd.tsx
norling: I like having a defined type, but I'll set it to `any` for this part of development and fix it once we have the first testing feedback from the users... Or is that lazy?

danr (approved):

PR #70, opened 2020-06-01
Update node to 14.3 in dockerfiles

danr:
This updates yarn which in turn gets rid of the error messages "the engine
'parcel' appears to be invalid".


airenzp (approved):
With this version it works!

PR #71, opened 2020-06-02
Feature/permissions

norling:
Basics of how I understood them to want the permission system to work. The permission levels are based on the permission levels set in the database; 'public, 'authenticated', and 'private'. 

I am honestly not quite sure if this is how they intended permissions to propagate, but I figured it was easier to make something that they could test before having the discussion. If this looks good to you I'm gonna go ahead and add some forms for granting and revoking roles on the page, and we should be almost ready for testing!

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -206,7 +206,7 @@ class Herd(BaseModel):
                     del data['latitude']
                     del data['longitude']
                 else:
-                    target_field = field[:-8]
+                    target_field = field[:-len('_privacy')]
                     del data[target_field]
             # remove the access level value if the user doesn't have private
             # access
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -429,7 +429,9 @@ def insert_data(filename='default_data.json'):
     }
     ```
     """
-    data = json.load(open(filename))
+    infile = open(filename)
+    data = json.load(infile)
+    infile.close()
 
     for table, values in data.items():
         if table not in [m.__name__ for m in MODELS]:
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -8,6 +8,7 @@ import uuid
 import logging
 
 from peewee import (PostgresqlDatabase,
+                    SqliteDatabase,
                     Proxy,
                     Model,
                     DoesNotExist,
@@ -31,17 +32,20 @@ from werkzeug.security import (
 DB_PROXY = Proxy()
 DATABASE = None
 
-def set_database(name, host, port, user, password):
+def set_database(name, host=None, port=None, user=None, password=None, test=False):
     """
     This function makes it possible to set the database manually when settings
     aren't loaded.
     """
     global DATABASE #pylint: disable=global-statement
-    DATABASE = PostgresqlDatabase(name,
-                                  host=host,
-                                  port=port,
-                                  user=user,
-                                  password=password)
+    if test:
+        DATABASE = SqliteDatabase(name)
+    else:
+        DATABASE = PostgresqlDatabase(name,
+                                    host=host,
+                                    port=port,
+                                    user=user,
+                                    password=password)
     DB_PROXY.initialize(DATABASE)
 
 try:
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -22,9 +22,6 @@ from peewee import (PostgresqlDatabase,
                     TextField,
                     UUIDField,
                     )
-from playhouse.postgres_ext import (
-    JSONField
-)
 
 from werkzeug.security import (
     check_password_hash,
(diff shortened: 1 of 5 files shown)


app/utils/database.py
```diff
@@ -286,7 +296,54 @@ class User(BaseModel):
     uuid = UUIDField()
     password_hash = CharField(128)
     validated = BooleanField(default=False)
-    privileges = TextField(null=True)
+    _privileges = TextField(column_name="privileges", default='{"roles":[]}')
```
danr: I found this naming scheme and the database usage interesting. Let's look at it.
First, what other privs than `roles` are you expecting?
Secondly, you store it as a json string in the DB. I don't have a particularily strong objection about this but how come you aren't storing in a row-oriented/SQL-fashion instead?

app/utils/database.py
```diff
@@ -296,6 +353,48 @@ def frontend_data(self):
...
 
+    def genebank_permission(self, genebank_id):
+        """
+        Returns the permission level for the genebank given by `genebank_id`.
+        Permission levels are set as 'public', 'authenticated', 'private'.
+        """
+
+        access_level = 'public'
+        for role in self.privileges['roles']:
+            if role['role'] == 'admin':
```
danr: The way you are using the name role now is for two things:
```typescript
type role₁ = {role₂: role₂, target: string]
type role₂ = 'admin' | 'specialist' | 'manager' | 'owner'
```
In addition, privileges are only roles:
```typsecript
type privileges = {roles₁: role₁[]}
```
This dual naming is apparent in this code which with the indexes would be
```python
for role₁ in self.privileges['roles₁']:
    if role₁['role₂'] == 'admin':
         ...
```
I don't know what you're after but one way to simplify this assuming that there are no other privs than roles would be:
```typescript
type privileges = privilege[]
type privilege = { role: role, target: string }
type role = 'admin' | 'specialist' | 'manager' | 'owner'
```
A less invasive change would be to change `role₁` to `targeted_role`:
```typescript
type privileges = {targeted_roles: targeted_role[]}
type targeted_role = {role: role, target: string]
type role = 'admin' | 'specialist' | 'manager' | 'owner'
```

danr (changes requested):
The role naming should be improved and preferably some comment on why json is stored in the database should be added.

app/utils/database.py
norling: When we talked about it in chat this was the solution people seemed to favor (while I was slightly more in of having permissions in tables). The privileges field was already present in the database, so I used it... :)

pontus:
This has somewhat more roles than I expected, I'm not sure if I'm missing something, but I thought we'd essentially have genebank manager, herd manager, authenticated user (and anonymous, if one consider that a role).

pontus:
I'm uncertain about the approach with access levels as something computed from the roles rather than simply having binary functions to respond whatever access should be permitted.

I find the introduction of privileges mostly confusing.

norling:
About the roles; these are the roles that has been requested. Admin (from jonas, to be used by system administrator), genebank manager (to manage genebank), genetic specialist (to evaluate genebank data, but shouldn't have editing rights like the manager has), and herd owner. I don't think we have any way of reducing this number.

danr:
The better names clears it up a bit. Now I'm confused by the difference between specialist and manager (but I have also not paid attention to the discussion before this).

Continuing my nitpicking though, you could now change the type of roles to
```typescript
type Role = 
  | { level: 'admin' }
  | { level: 'owner', herd: id }
  | { level: 'specialist' | 'manager', genebank: id }
```

app/utils/database.py
```diff
@@ -296,18 +296,26 @@ class User(BaseModel):
...
 
     @property
     def privileges(self):
         """
-        Wrapper property to convert a json text string to a python object
+        Wrapper property to convert a json text string to a python object.
+        The privileges are a list of roles, formatted as:
+        [
+            {'level': 'admin' | 'specialist' | 'manager' | 'owner',
+             'target_id: None        (if level == 'admin') |
```
danr: Missing `'` in `'target_id'`

app/utils/database.py
```diff
@@ -296,18 +296,26 @@ class User(BaseModel):
...
+                         hard_id     (if level == 'owner') |
```
danr: `herd_id` typo

danr (dismissed):
typos

norling:
and about the access levels - that was mostly a way of trying to generalize a bit, and the access levels were already present in the database for the restricted herd fields, so I figured it would make the system more uniform to keep using that. Would you prefer to have a separate system that just deals with the roles and their access?

app/utils/database.py
```diff
@@ -158,6 +171,50 @@ class Herd(BaseModel):
...
+        # prune data according to access level
+        for field in [f for f in data.keys() if f.endswith('_privacy')]:
+            # remove values if access_level is less than required
+            field_level = data[field] or 'private'
+            if levels.index(access_level) < levels.index(field_level):
+                if field == 'coordinates_privacy':
+                    del data['latitude']
+                    del data['longitude']
+                else:
+                    target_field = field[:-8]
```
danr: More apparent what the `-8` does by explicitly writing the expression it came from: 
```python
target_field = field[:-len('_privacy')]
```

danr:
Me and Martin will philosophise a bit about how to rewrite this for easier testing tomorrow Thursday, June 4 on zoom

app/utils/database.py
```diff
@@ -148,6 +171,50 @@ class Herd(BaseModel):
...
+                    del data[target_field]
+            # remove the access level value if the user doesn't have private
+            # access
+            if access_level != levels.index('private'):
```
pontus: Already tested by the handling at L204 so this should go away?

norling: no. line 204 checks that you have permission to access the value field, which can vary in required permission, while this line removes the permission field which should only be accessible if you have specifically private permissions.
like, a field could have permission level "authenticated", and that could pass the check on line 204, but wouldn't pass 2013.

pontus: I don't think I understand the logic then, but in addition: this compares the access_level (string) to the index in the list (integer).

But with a straight comparison if access_level is private, it would remove fields marked as public (that thus should be available to everybody)?

danr: This is fixed by using string comparison in #73

pontus (approved):
Ok

PR #72, opened 2020-06-02
Add module for computing inbreeding coefficients

jonandernovella:
A PoC module has been added in order to compute inbreeding coefficients using pydigree.

### DIFF
--- a/app/utils/inbreeding.py
+++ b/app/utils/inbreeding.py
@@ -1,18 +1,23 @@
 """
 Inbreeding coefficent calculations from the Herdbook's database records.
-The inbreeding coefficient is the Malecot coefficient of coancestry of the two parents of an individual.
+The Malecot coefficient of coancestry of the two parents is calculated.
 """
 
 from pydigree.population import Population
-from pydigree.individual import Individual
 from pydigree.pedigree import Pedigree
 from pydigree.io.base import PEDRecord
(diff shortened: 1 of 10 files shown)


app/utils/inbreeding.py
```diff
@@ -0,0 +1,96 @@
...
+class IndividualPEDRecord(PEDRecord):
+    def __init__(self, individual):
+        """
+        Creates pedigree record from a dictionary of an individual
+
+        :param individual: A dictionary containing an individuals info
+        :type individual: dict
+        """
+        self.ind_id = individual["id"]
+        self.fam = individual["nerd_id"]
```
danr: herd id?

danr: :nerd_face:

norling (changes requested):
Looks good in general, but there are some minor things:

1. `/api/coefficients` is not a great name for the endpoint. Maybe something like `/api/inbreeding`?
2. There are some pylint complaints, and it's always nice to keep pylint happy! :) 
3. Could you add some tests just to make sure that things only crash and burn when appropriate? 
4. There should probably be some access control on this endpoint, so just check the target individuals genebank against the `user.accessible_genebanks` (once #71 is merged)

norling (approved):

PR #73, opened 2020-06-04
Feature/remove fields by privacy

danr:
:v:


norling (approved):
requirements.txt should use fixed versions, but we can fix that later.

PR #75, opened 2020-06-05
Add specific colour codes for Mellerud

andkaha:
Mellerud rabbits can only ever be black/white or albino.
Use colour codes 100 and 101 for these.


norling (approved):
Looks good!

PR #79, opened 2020-06-12
Add fields to the herd table

andkaha:
* herd.herd_name:	The herd name (herd.name is a person's name).
* herd.is_active:	Whether the herd is currently active or not.
* herd.start_date:	When the herd was first established/registered.

All these fields may be NULL for "unknown".


norling (approved):
Looks very reasonable

PR #82, opened 2020-06-23
Feature/frontend layout

norling:
It's not pretty, but I'm hoping that it's enough functionality to test during the summer! :)


app/utils/database.py
```diff
@@ -462,6 +531,30 @@ def is_admin(self):
         """
         return 'admin' in [p['level'] for p in self.privileges]
 
+    @property
+    def is_manager(self):
```
pontus, line 547: Can we get a little reasoning (in PR is fine) about the naming convention here; why is it is_manager (which implies booleanity to me) and not genebanks_managed (or preferably a better name)?

frontend/genebanks.tsx
```diff
@@ -1,33 +1,98 @@
...
+    textDecoration: "none",
+  },
+  spacer: {
+    height: '39px',
+  }
+});
+
+import * as ui from './ui_utils'
+
+function InnerPaper(props: {id: number}) {
```
danr, line 33: Oops I had hoped that you would find a more descriptive name than `InnerPaper` here :D

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  return <>
+    {loading && <h2>Loading...</h2> ||  herd &&
+      <>
+        <h1>{herd.name || `Besättning ${herd.id}`}</h1>
+        <form className={classes.form} noValidate autoComplete="off">
+          {simpleFields.map((field: any, i: number) => {
+            return <TextField
+                    key={i}
+                    id={field.key}
```
danr: Rename key to id then ;)

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }, [props])
+
+  const setFormField = (label: string, value: string | number) => {
+    herd[label] = value;
+    setHerd(herd);
+  }
+
+  const submitForm = () => {
+    let postData = Object.assign({}, herd);
```
danr: `const postData = {...herd}`

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+  simpleField: {
+    width: "400px",
+  },
+});
+
+/**
+ * Provides herd management forms for setting herd metadata.
+ */
+export function HerdForm(props: {id: any | undefined}) {
```
danr: What's with the type? `any | undefined` = `any`. Did you mean `string | undefined`?

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+  const [herd, setHerd] = React.useState(undefined as any)
```
danr: You have types for herds, why not use `undefined | Herd`?

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+          setLoading(false);
+        },
+        error => console.error(error)
+      );
+    }
+  }, [props])
+
+  const setFormField = (label: string, value: string | number) => {
+    herd[label] = value;
+    setHerd(herd);
```
danr: `herd && setHerd({...herd, [label]: value})`

danr: Also you can type `const setFormField = <K extends keyof Herd>(label: K, value: Herd[K]) => ...` :nerd_face:

frontend/herdForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+                     onChange={e => setFormField('longitude', e.target.value)} />
+        </form>
+        <Button variant="contained"
+                color="primary"
+                onClick={() => submitForm()}>
+          Spara
+        </Button>
+        <h2>Individer</h2>
+        <ul>
+          {herd.individuals.map((individual: any, i:number) => {
```
danr: Hmm `Herd`s have no `individuals`... Is this actually a `Genebank`? :thinking:

frontend/individuals.tsx
```diff
@@ -0,0 +1,96 @@
...
+  {field: 'mother', title: 'Moder'},
+  {field: 'father', title: 'Fader'},
+  {field: 'color', title: 'Färg'},
+  {field: 'colorNote', title: 'Färganteckning'},
+]
+
+/**
+ * Shows a list of all genebanks, with links to the individual genebanks.
+ */
+export function IndividualsTable({id}: {field: number}) {
```
danr, line 87: Type should be `{id: number}`?

frontend/manage_users.tsx
```diff
@@ -0,0 +1,81 @@
...
+  return <>
+    <div className={classes.sidebar}>
+    <Tabs
+        orientation="vertical"
+        variant="scrollable"
+        value={userTab}
+        onChange={userChange}
+        className={classes.verticalTabs}
+      >
+      {users.map((u:any, i:number) => <Tab key={i} label={u.email} />)}
```
danr: u is a `NameID` which only has a `name` and `id` field, no `email` field

frontend/manage_users.tsx
```diff
@@ -0,0 +1,81 @@
...
+  centerButton: {
+    width: "80%",
+  }
+});
+
+export function ManageUsers() {
+  const classes = useStyles()
+  const {users} = useDataContext()
+  const [userTab, setUserTab] = React.useState(0);
+  const [selectedUser, selectUser] = React.useState(undefined as number | string | undefined)
```
danr, line 45: This should probably not be called `selectedUser` but something with edit action.

You can narrow down the string type to `'new'`: `number | 'new' | undefined`

frontend/manage_user.tsx
```diff
@@ -0,0 +1,220 @@
...
+          default: console.debug("error:", data)// "failed" or other error
+        }
+      },
+      error => console.error(error)
+    )
+  }
+
+  return <>
+    {user && <>
+      <h2>{isNew ? 'Ny användare' : user.email}</h2>
```
danr, line 121: Move this and its state to an own `CreateNewUser` component

frontend/manage_user.tsx
```diff
@@ -0,0 +1,220 @@
...
+  email: string
+  validated: boolean
+  privileges: any
+  password?: string
+}
+
+/**
+ * Provides a form for changing user metadata and user roles.
+ */
+export function ManageUser({id}: {id: number | string | undefined}) {
```
danr, line 37: Here `id: number | 'new' | undefined` should probably not be called id...

Maybe split this into two components `ManageUser` and `CreateNewUser` and call the correct one from `ManageUsers`?

frontend/manage_user.tsx
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }, [genebanks])
+
+  const selectGenebank = (gId: number) => {
+    setGenebank(gId);
+    let genebankData = genebanks.filter(g => g.id == gId);
+    if (genebankData.length > 0) {
+      setHerds(genebankData[0].herds)
+      if (herds.length > 0 && herds.filter(h => herd == h.id).length == 0) {
+        setHerd(herds[0].id)
```
danr, line 76: Why do you first filter and then anyway set to the first elemnet?

frontend/manage_user.tsx
```diff
@@ -0,0 +1,220 @@
...
+                     type='email'
+                     className={classes.simpleField}
+                     value={user.email ?? undefined}
+                     onChange={e => setUser({...user, email: e.target.value})}
+            />
+          {isNew ? <TextField label='Lösenord'
+                      type='password'
+                      hidden={true}
+                      className={classes.simpleField}
+                      value={user.password ?? undefined}
```
danr, line 133: This needs to be `''` instead of `undefined` otherwise React gets confused if this is a controlled or uncontrolled component.

frontend/manage_users.tsx
```diff
@@ -0,0 +1,81 @@
...
+import { Tabs, Tab } from '@material-ui/core/'
+import { Box, Button } from '@material-ui/core'
+
+import { useDataContext } from './data_context'
+import { ManageUser } from '~manage_user';
+
+// Define styles for tab menu
+const useStyles = makeStyles({
+  sidebar: {
+    float: "left",
```
danr, line 17: Why not put sidebar and controls side-by-side using flex instead of float?

danr: Then you don't need to do all the calc.

frontend/individuals.tsx
```diff
@@ -0,0 +1,96 @@
...
+  }, [genebanks])
+
+  return <>
+    <div className={classes.table}>
+      <MaterialTable
+        icons={tableIcons}
+        columns={columns}
+        data={individuals}
+        title="Alla individer"
+      />
```
danr, line 106: Moder och Fader seems to refer to some internal id (like 11793) while Nummer for a given bunny is written in another form (like G1390-1911).

frontend/manage.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Provides genebanks management forms for granting and revoking herd
+ * permissions, and managing herd animals.
+ */
+export function Manage() {
+  const {genebanks} = useDataContext()
+
+  const tabs: ui.RoutedTab[] = genebanks.map(g => ({
+    path: `/${g.id}`,
```
danr, line 86: What about 
```typescript
    path: `/${g.name.toLowerCase().replace(/åä/g, 'a').replace(/ö/g, 'o')}`,
```
?

frontend/explore_herds.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Box, Button } from '@material-ui/core'
+import { Herd } from '~herd'
+import { Genebank } from 'data_context_global'
+
+import { useDataContext } from './data_context'
+
+// Define styles for tab menu
+const useStyles = makeStyles({
+  sidebar: {
+    float: "left",
```
danr, line 17: Some interesting overflow is happening here
image

frontend/individuals.tsx
```diff
@@ -0,0 +1,96 @@
...
+// Define styles for tab menu
+const useStyles = makeStyles({
+  table: {
+    height: "calc(100% - 75px)",
+    padding: "0 20px",
+    overflowY: "scroll",
+  },
+});
+
+const tableIcons = {
```
danr: Move all the icons stuff to an own file?

frontend/manage_herds.tsx
```diff
@@ -0,0 +1,96 @@
...
+    width: "calc(100% - 250px)",
+    padding: "0.5cm 1cm",
+    overflowY: "scroll",
+  },
+  centerButton: {
+    width: "80%",
+  }
+});
+
+export function ManageHerds({id}: {id: number | undefined}) {
```
danr, line 42: There is a lot duplicated here and in `ExploreHerds`. Why not just add an edit button when Exploring if you're an admin?

frontend/data_context_global.tsx
```diff
@@ -0,0 +1,84 @@
...
+    email_verified: boolean | null
+    www: string | null
+    www_privacy?: string | null
+    mobile_phone: string | null
+    mobile_phone_privacy?: string | null
+    wire_phone: string | null
+    wire_phone_privacy?: string | null
+    latitude: string | null
+    longitude: string | null
+    coordinates_privacy?: string | null
```
danr, line 60: Many fields here are both `undefine`-able (with `?`) and `| null`. Why both?

danr (changes requested):
Comments from a (first?) run through the frontend. Also some accompanying changes in #83

app/utils/database.py
norling: yeah... it's named that way since it used to be a boolean ^^ Then I changed it to give extra info. I should separate the two though!

frontend/herdForm.tsx
norling: I don't mind this syntax, but you can change to spread if you like ^^

frontend/data_context_global.tsx
norling: they're undefined when they're not present and null when they're unset.

frontend/manage_herds.tsx
norling: yeah - it wasn't originally supposed to be so similar, but I ran out of time. so I'll have to think about the final design before I decide if there's value in optimizing.

frontend/herdForm.tsx
norling: yeah, that was the plan - I just ran out of time, so the old code never got the new types. I'll have a look now while we're looking!

frontend/herdForm.tsx
norling: I prefer it as 'key' or 'name' as it's a reference to key->value pairs of the herd object. I can remove the id tag though.

frontend/herdForm.tsx
norling: haha - I think I meant string | number | undefined, but got confused ^^

frontend/herdForm.tsx
norling: haha - I considered something like that but didn't get it to work, but I'll use this ^^

frontend/herdForm.tsx
norling: No, it's just a herd that didn't have the Herd type

frontend/herdForm.tsx
```diff
@@ -51,9 +52,8 @@ export function HerdForm(props: {id: string | number | undefined}) {
     }
   }, [props])
 
-  const setFormField = (label: string, value: string | number) => {
-    herd[label] = value;
-    setHerd(herd);
+  const setFormField = <K extends keyof Herd>(label: K, value: Herd[K]) => {
+    herd && {...herd, label: value}
```
danr: You lost `setHerd` :fearful:

frontend/data_context_global.tsx
danr: Ah I get it, it's for the `_privacy` fields

frontend/herdForm.tsx
danr: I see! I got confused thinking it would be react keys.

frontend/manage_user.tsx
norling: it might be a dumb way to write this check, but it's essentially "if there are herds, and the currently selected herd isn't in the current list of herds, then select the first herd as default" could be `herds.length > 0 && !herds.find(h => h.id == herd)` instead, and that might be nicer...

frontend/herdForm.tsx
danr: OK! :grimacing: 

But please use `const` when you don't need `let`!

frontend/manage_user.tsx
norling: ok! good to know! :)

frontend/herdForm.tsx
danr: And the brackets in `[label]`!

`setHerd({...herd, [label]: value})`

frontend/manage_users.tsx
norling: I honestly don't remember. I almost always use flex, so I want to think I had a reason for a float, but I can't remember what it would be :D I'll have a look through the rules once we have some more data on what they want from the thing though! I'll probably design everything like a big flex-thing in the end...

frontend/individuals.tsx
norling: yeah - this is purely a database optimization issue. I'd prefer to have parents as NameID or something like that, but I haven't had time to sort out a fast enough database query to get all the info I want.

frontend/herdForm.tsx
norling: haha - yeah, that was dumb ^^

frontend/manage_user.tsx
danr: Haha okej now I see it. I overlooked the `== 0` part but now that you pointed it out it's obvious of course. Yes `herds.length && !herd.find(...)` would be clearer. (Too bad `!` isn't `not` in JS though!)

frontend/manage_users.tsx
danr: Alright!

frontend/individuals.tsx
danr: Alright!

frontend/manage_herds.tsx
danr: One option is just to write this down in an issue and let it be like it is right now of course.

frontend/herd_view.tsx
```diff
@@ -28,6 +38,14 @@ export function HerdView({id}: {id: string}) {
     {herd &&
       <>
         <h2>{`G${herd.herd}`} {herd.herd_name ? `- ${herd.herd_name}` : ''}</h2>
+        <dl>
+          {fields.map(field => <>
```
danr: It's only after `return` you need `<> ... </>` or `( ... )`, not when you're in an unfinished expression such as to the right of `=>`.

danr: Oh I didn't see you had both `<dt>` and `<dd>`, then you need a `<>`-fragment of course.

norling: yeah, but I had to change it to a div to get a key on it, so I still did it wrong ^^

jhagberg:
The Mellerud data is not listing 

/genebanks/2/individuals
and 
/genebanks/1/individuals

is the same list

frontend/genebanks.tsx
norling: it works for now ^^

frontend/manage.tsx
norling: links will probably be changed to the G-id later, so I'll keep this for now.

frontend/individuals.tsx
norling: Yeah - we can do that once the table is finished. In the mean time I'd prefer to keep them here so that I can look and change easier :)

danr (approved):

PR #85, opened 2020-06-25
useMartinState PoC

danr:
I got feeling and had to try this out :nerd_face: 

Not exactly recommending that we use this now but you know, just wanted to see how it looked like.


norling (commented):
Haha - känns mest som ett bökigt sätt att använda react ^^

PR #91, opened 2020-07-21
Feature/inbreeding

airenzp:
This branch adds mainly the inbreeding coefficient to the interface, with cache management. It also includes a bugfix to avoid errors when one of the parents is missing, but not both. 
The documentation for the dev setup is also updated with the steps required to run a local environment and create an admin user.


docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
             - POSTGRES_HOST_AUTH_METHOD=trust
         volumes:
             - ./postgres-data:/var/lib/postgresql/data
+        ports:
+            - 5432:5432
```
pontus: Why do we want to expose the database?

README.md
```diff
@@ -26,11 +26,29 @@ breeding/preservation effort for a race.
...
+5c9193fc774c        herdbook_frontend:latest   "docker-entrypoint.s…"   20 hours ago        Up 3 hours                                                                                   theherdbook_herdbook-frontend-devel_1
+
+```
+
+To access the local server deployed open this url https://localhost:4443 in your browser. You will need to configure the browser to allow self-signed localhost certificates. In Chrome, this can be done by accessing this property from the browser: `chrome://flags/#allow-insecure-localhost` and setting its vale to Enabled. In Firefox, when loading the url you can click on Advanced and  add the exception suggested from the browser.
+
+To be able to login on the website and play with it you will need to create an user with admin privileges. This can be done by executing the following command from the app folder, using your email and password:
+
+```
+python3 -c "from utils.data_access import register_user; register_user('user@domain.com', 'userpassword', privileges=[{'level': 'admin'}])
```
pontus: Better to refer to `register_user.sh` here (and adding privileges there).

jonandernovella (approved):

frontend/individual.tsx
```diff
@@ -60,6 +60,7 @@ export function Individual({id}: {id: string}) {
                   : '-'
                 }
               </dd>
+            <dt>Navelskoefficient</dt> <dd>{individual.inbreeding}%</dd>
```
danr: Isn't it _Inavelskoefficent_?

danr (commented):
One typo

app/herdbook.py has some formatting oddities... run black?

frontend/individual.tsx
jhagberg: yes

airenzp: tack! I edited the file

danr (approved):

pontus (changes requested):

norling (approved):

docker-compose.yml
airenzp: ok, I remove it. Was using it in my local env

README.md
airenzp: register user was trying to run on localhost, and failing, in this way I could solve it

airenzp: I tried the script again and it is working. I refer now to the script and add the admin rights there.

danr:
What about running black with `--skip-string-normalization`? That way it won't touch all the single quotes.

It is good to have a standard, it does not really bother me. What ever you prefer, but I already pushed this version. Seems unnecessary.

pontus (approved):

PR #96, opened 2020-08-20
Change herd.herd to VARCHAR(10) from INTEGER

andkaha:
This allows us to store the herd IDs with a "G" or "M" prefix in the
database directly.

### DIFF
--- a/app/requirements.txt
+++ b/app/requirements.txt
@@ -9,7 +9,7 @@ MarkupSafe==1.1.1
 mccabe==0.6.1
 peewee==3.13.2
 psycopg2-binary==2.8.5
-git+git://github.com/jameshicks/pydigree@master
+#git+git://github.com/jameshicks/pydigree@master
 pylint==2.4.4
 six==1.14.0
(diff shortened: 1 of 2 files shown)


app/requirements.txt
```diff
@@ -9,7 +9,7 @@ MarkupSafe==1.1.1
 mccabe==0.6.1
 peewee==3.13.2
 psycopg2-binary==2.8.5
-git+git://github.com/jameshicks/pydigree@master
+#git+git://github.com/jameshicks/pydigree@master
```
jhagberg: Why do you comment this line is not related to the database changes?

jhagberg (changes requested):

app/requirements.txt
andkaha: This was not supposed to be committed. It was a local change that I needed to be able to load the data (the `pydigree` Python module did not install on my system).  I will revert this change.

norling (approved):
Looks reasonable!

jhagberg (approved):
LGTM

PR #98, opened 2020-08-25
Feature/pedigree graph

airenzp:
This branch adds the pedigree graph to the individuals. The pedigree is built according to the number of generations specified, which by default is 5. One can navigate through the nodes in the tree using double click and use the navigation buttons for zoom and move. A herd pedigree can also be rendered accessing the herd view and url.
In order to avoid compilation issues, the volumes were removed from the frontend docker, since they are already added in the docker-compose file, and the paths updated.


.docker/frontend
```diff
@@ -5,11 +5,11 @@ COPY /frontend/ /app/
 WORKDIR /app
 RUN yarn install
 
-VOLUME /app/node_modules
-VOLUME /app/.cache
-VOLUME /app/.parcel-cache
+#VOLUME /app/node_modules
+#VOLUME /app/.cache
+#VOLUME /app/.parcel-cache
```
danr: Why were these removed?

app/utils/data_access.py
```diff
@@ -415,6 +415,56 @@ def get_individuals(genebank_id, user_uuid=None):
         return []
 
 
+
+def get_leafs(genebank_id, user_uuid=None):
```
danr: Typo? `get_leaves`

frontend/genebank_pedigree.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+  React.useEffect(() => {
+
+    get(`/api/genebank_pedigree/${id}`).then(
+      data => data && setPedigree(data),
+      error => console.error(error)
+    )
+  }, [id])
+
+  class GenebankPedigreeNetwork extends Component {
```
danr: This is written in old-style class-based React which is not used elsewhere in the project.

frontend/genebank_pedigree.tsx
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    componentDidMount() {
+      this.network = new Network(this.appRef.current, pedigree, this.options);
+      this.network.on("doubleClick", this.onNodeClick)
+    }
+
+    onNodeClick(params) {
+      if (params.nodes.length > 0) {
+        var nodeid = params.nodes[0];
```
danr: Prefer `const` (or `let`) over `var`

frontend/genebank_pedigree.tsx
```diff
@@ -0,0 +1,99 @@
...
+        window.location.replace("/pedigree/" + nodeid);
```
danr: There must be some way to change this without having to go through a full page reload?

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,160 @@
...
+      data => data && setIndividual(data),
+      error => console.error(error)
+    )
+    get(`/api/pedigree/${id}/${generations_input}`).then(
+      data => data && setPedigree(data),
+      error => console.error(error)
+    )
+  }, [id, generations_input])
+
+  class PedigreeNetwork extends Component {
```
danr: Again old-style class-based component

danr: This looks with a brief glance like an exact copy of the class from the other file... Perhaps some of the code can be shared?

frontend/package.json
```diff
@@ -23,20 +23,27 @@
...
   },
   "scripts": {
-    "start": "parcel serve index.html --port 2345 --hmr-port 2346 --no-source-maps",
-    "build": "rm dist/* -rf; parcel build index.html --no-source-maps",
+    "start": "parcel serve index.html --port 2345 --hmr-port 2346",
+    "build": "rm dist/*; parcel build index.html",
     "test": "jest"
   },
   "dependencies": {
+    "@babel/plugin-transform-flow-strip-types": "^7.10.4",
```
danr: Why is this added? There are no flow types are there?

frontend/package.json
```diff
@@ -23,20 +23,27 @@
...
     "@material-ui/core": "^4.10.1",
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.50",
+    "@parcel/transformer-react-refresh-babel": "^2.0.0-alpha.3",
```
danr: Hmm I never needed to add this package explicitly to package.json

frontend/package.json
```diff
@@ -23,20 +23,27 @@
...
+    "@parcel/watcher": "^2.0.0-alpha.8",
```
danr: Nor this

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,160 @@
...
+      return (
+        <div ref={this.appRef}  >
+        </div>
+      );
+    }
+  }
+
+  return <>
+    {individual && pedigree && <>
+      <table width="100%">
```
danr: This table is copied from some other file, isn't it?

danr (commented):
Some comments

docker-compose.yml
```diff
@@ -13,6 +13,8 @@ services:
             - POSTGRES_HOST_AUTH_METHOD=trust
         volumes:
             - ./postgres-data:/var/lib/postgresql/data
+        ports:
+            - 5432:5432
```
pontus: I think we still don't want to expose the database.

docker-compose.yml
```diff
@@ -27,11 +29,11 @@ services:
...
             - type: bind
               source: ./frontend
-              target: /frontend
+              target: /app
             - /frontend/node_modules
             - /frontend/.parcel-cache
             - /frontend/.cache
-        working_dir: /frontend
+        working_dir: /app
     main:
```
pontus: As discussed in slack, this will pollute the source directory, please revert/remove this from PR.

.docker/frontend
airenzp: one can create volumes  using docker-compose without declaring them in the Dockerfile. There was also an error with the paths, that referred to a frontend folder that did not exist in the dockerfile but did in the docker-compose. I commented this before with Pontus in Slack...Hope this makes it clearer.

app/utils/data_access.py
airenzp: yes, I will fix it, sorry.

frontend/genebank_pedigree.tsx
airenzp: Mmmm, ok. This is my first react app. I will check it.

frontend/genebank_pedigree.tsx
airenzp: ok

frontend/genebank_pedigree.tsx
airenzp: ok, will check it

frontend/individual_pedigree.tsx
airenzp: ok

airenzp: ok

frontend/package.json
airenzp: probably can be removed. I was trying to solve my compilation issues at first installing what it seemed missing libraries, according to the error. I will check it.

frontend/package.json
airenzp: same here

frontend/package.json
airenzp: same here, as in my previous comment, will check it

frontend/individual_pedigree.tsx
airenzp: what do yo mean?

frontend/genebank_pedigree.tsx
norling: yeah, try letting react do the routing with `react-router-dom`. You can import `useHistory` and then do `history.push` to navigate. This can be useful if you wish to leave a "trail" so that people can go back through their changes by navigating back. If you just want to update you can use the functions in the `communication.tsx` file to fetch data asynchronously!

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,160 @@
...
+                </dd>
+
+              </dl>
+            </td>
+            <td width="85%" >
+              <PedigreeNetwork />
+            </td>
+          </tr>
+        </tbody>
+      </table>
```
norling: Since we're using material-ui for the styles in most of the app - consider using their tables ([here](https://material-ui.com/components/tables/)) to get a consistent look on the page easily! :)

app/herdbook.py
```diff
@@ -9,16 +9,23 @@
...
 )
-from flask_caching import Cache
+
+from werkzeug.urls import url_parse
 import utils.database as db
 import utils.data_access as da
 import utils.inbreeding as ibc
+import logging
+from flask_caching import Cache
+from flask_login import login_required, login_user, logout_user, current_user, LoginManager
```
norling, line 28: I don't mind using the Flask login system, but I'm not a big fan of using different systems on different endpoints. I seem to remember that the flask login system didn't really play that well with asynchronous requests, but if you can get it to work - please change all the endpoints so that the app remains consistent! :)

norling (changes requested):
Some comments that are mostly up to you, but I would really like if you'd look over the login system before we merge!

app/herdbook.py
airenzp: I did not touch your login system, just added what I needed to use the login_required attribute on the python methods. This does not change the react form or any logic behind. But I would like to be able to implement the return to an url after login, for example. For that, some integration with react needs to be done, since the return url is sent to the login page

frontend/individual_pedigree.tsx
airenzp: ok, thanks. I will take a look at that.

frontend/genebank_pedigree.tsx
airenzp: thanks, ok

app/herdbook.py
norling: Absolutely! But if we use decorators for one endpoint we should use them for all endpoints that require login so that the system remains consistent. Adding a login-redirector sounds like a great plan though! :)

airenzp: Sure. I just did not wanted to do that without your approval

frontend/individual_pedigree.tsx
airenzp: I used the individual.tsx file as a starting point and added the pedigree graph.

app/utils/data_access.py
airenzp: solved

frontend/genebank_pedigree.tsx
airenzp: solved

frontend/genebank_pedigree.tsx
airenzp: used history

frontend/package.json
airenzp: solved

frontend/package.json
airenzp: solved

frontend/package.json
airenzp: solved

frontend/genebank_pedigree.tsx
airenzp: solved

frontend/individual_pedigree.tsx
airenzp: solved

airenzp: solved

frontend/individual_pedigree.tsx
airenzp: should I remove the previous individual.tsx?

frontend/individual_pedigree.tsx
airenzp: solved

app/herdbook.py
airenzp: done

pontus (changes requested):
Nice work

docker-compose.yml
airenzp: ok

docker-compose.yml
airenzp: As I said before, these volumes are already in the develop branch, they are not my creation. I just fixed the path. If we do not want them, then Dan needs to approve this, as it is him who added them. I do not know the effects of removing the volumes.

airenzp: @danr you might want to comment your ideas on this matter.

frontend/individual_pedigree.tsx
danr: I don't know, up to you and Martin :)

frontend/genebank_pedigree.tsx
danr: Nice! :+1:

frontend/genebank_pedigree.tsx
danr: Perfect! :+1:

docker-compose.yml
danr: I don't know, I'm very confused about the docker volumes for the cache directories. If you're happy with this setup and don't run into caching problems, go for it!

frontend/pedigree.tsx
```diff
@@ -0,0 +1,58 @@
...
+    () => {
+      network.current = new Network(domNode.current, pedigree, options);
+      network.current.on("doubleClick", (params) => {
+          if (params.nodes.length > 0) {
+            const nodeid = params.nodes[0];
+            history.push("/individual/" + nodeid);
+          }
+      });
+    },
+    [domNode, network, pedigree, options]);
```
danr: Note that refs from useRef don't ever change so these don't need to be listed in the dependency list, see https://reactjs.org/docs/hooks-reference.html#useref
I think the reason why this still works is because the domNode ref is updated before your callback is run. One alternative would be to put the ref in a state instead, as in https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node

You don't need to list options in the dependency list either since it is a constant reference (declared with `const`!)

frontend/pedigree.tsx
```diff
@@ -0,0 +1,58 @@
+import React, { Component, useRef, useEffect } from 'react'
+import { Network } from 'vis-network';
+import "vis-network/styles/vis-network.css"
+import { useHistory } from "react-router-dom";
+
+
+const PedigreeNetwork = ({pedigree}) => {
```
danr: I don't know how much "typing coverage" you're aiming for in this project but adding types here can be done like so:

```typescript
const PedigreeNetwork = ({pedigree}: {pedigree: string}) => {...
```

Also if you define components with `function` you see that name in debugging views (profiler, componentsv iew and error messages).

frontend/pedigree.tsx
```diff
@@ -0,0 +1,58 @@
...
+    },
+    [domNode, network, pedigree, options]);
+
+    return (
+        <div ref={domNode}  >
+        </div>
+    );
+};
+
+export default PedigreeNetwork;
```
danr: I don't usually use default exports but if you want to use it you can export it default in one go

```typescript
export default function PedigreeNetwork(...) { 
    ...
}
```

danr (approved):

docker-compose.yml
pontus: As I said in slack, what you do here is essentially removing the use of modules - they are still mounted under /frontend/node_modules etc. Whereas things are now run under /app (which will lead to the source directory being polluted by the cache folders and so on).

If you want to change to `/app`, I'd really want the volumes to be mounted there as well - but I suspect that will bring back your issue with the volume being corrupted over time (so I suspect you may want to run with the current setup in your specific environment).

frontend/pedigree.tsx
airenzp: Ok, all these concepts are still quite loose for me. But I am looking forward to properly study react! Nice to have someone to point in the right direction.

docker-compose.yml
airenzp: That might work, what was causing my compilation issues was having different volumen configurations in the dockerfile and the docker-compose. So you want to put everything inside the app folder instead? I am testing it now.

pontus: I'm fairly convinced that what you ran into was a bug for Docker on macOS (which quite possibly can come back if you start using the volumes again). I don't much care about the actual path used inside the container, but I do not want temporary files and node libraries to spill outside of the container.

airenzp: In the original code, there was this line in the dockerfile:
VOLUME /app/node_modules

And then in the docker-compose:
volumes:
            - type: bind
              source: ./frontend
              target: /frontend
              - /frontend/node_modules

You are saying in one place that your volume in the container is under the /app folder and in the other that the same volume is under the /frontend folder. Do we agree on this?

frontend/pedigree.tsx
danr: Seems like you're picking it up very quickly! Feel free to ask questions in the slack, there's the #frontend channel which has mainly been used to discuss React pecularities so far :)

docker-compose.yml
pontus: These would be independent (although possibly at some time in history intended to be the same, and that doesn't hurt but is not a strict requirement either).

The benefit of having the same path in both places is that the same volume is used (without going through to much trouble) and thus that the volume is "preloaded" on the first run of the instance. This means the `yarn install` and so on runs faster on the first launch.

Do note that for what happens in the container image, it's not unimportant whatever the `VOLUME` is used or not, having it will move what's under the given path to the new volume and thus it will not be included in the container image (it doesn't

That is not the same as the issue of changing some paths where things are mounted in the container but not other (e.g. changing the bind mount from `/frontend` to `/app` while still having the volumes at `/frontend/node_modules` and so on).

The `VOLUME` keyword in the Dockerfile will also register the volume for the container, so in theory, things should "just work" if one continue to use the same path. In practice, I've found that behaviour inconsistent, at least when having other mountpoints possibly interfering with it, so I prefer setting them up explicitly (which has been consistent in my experience).

airenzp: It seems like we are making some progress here.
Then, what do you want me to do?

pontus: Anything that won't spill the contents that did go in the volume over to the bind mounted source directory is fine, this could be:

* reverting this change
* keeping the change but also changing the volumes so they are mounted as subdirectories in the same directory as the source  is mounted

airenzp: like this?
volumes:
            - type: bind
              source: ./frontend
              target: /app
            - ./frontend/node_modules:/app/node_modules
            - ./frontend/.parcel-cache:/app/.parcel-cache
            - ./frontend/.cache:/app/.cache

pontus: IIRC that will be interpreted as additional bind mounts, which is what I don't want.

```
volumes:
  - type: bind
    source: ./frontend
    target: /app
  - /app/node_modules
  - /app/.parcel-cache
  - /app/.cache
```

should be fine.

airenzp: Great, it works. I have now pushed this version

pontus (approved):

frontend/pedigree.tsx
```diff
@@ -0,0 +1,62 @@
...
+    () => {
+      network.current = new Network(domNode.current, pedigree, options);
+      network.current.on("doubleClick", (params) => {
+          if (params.nodes.length > 0) {
+            const nodeid = params.nodes[0];
+            history.push("/individual/" + nodeid);
+          }
+      });
+    },
+    [domNode, network, pedigree]);
```
danr: You should still be able to remove the Refs from the dependency list (domNode and network)

airenzp: I just did it. Thanks

frontend/pedigree.tsx
airenzp: done

frontend/pedigree.tsx
airenzp: How can I see the frontend channel? I need to be added maybe?

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,104 @@
...
+      error => console.error(error)
+    )
+    get(`/api/pedigree/${id}/${generations_input}`).then(
+      data => mounted && data && setPedigree(data),
+      error => console.error(error)
+    )
+
+    return () => { mounted = false }
+
+  }, [generations_input])
```
danr: I would add `id` to the depedency list since there is a dependency.

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,104 @@
...
+import TableRow from '@material-ui/core/TableRow';
+
+
+/**
+ * Shows the information of a given individual and the pedigree graph built using the vis-network component
+ */
+export function IndividualPedigree({ id, generations }: { id: string }) {
+  const [pedigree, setPedigree] = React.useState(undefined as any)
+  const [individual, setIndividual] = React.useState(undefined as any)
+  const [generations_input, setGenerations] = React.useState(generations)
```
danr, line 18: Note that generations_input won't change if the generations prop (input) is changed (from parent).

One way to sync them up is to add an effect depending on `generations` that sets this state to `generations` when it changes.

airenzp: But the generations prop is not expected to change, this value is read from the url parameters, like the id in the previous comment. I thought this dependency was meant to be used for a changing parameter.

frontend/individual_pedigree.tsx
airenzp: ok

frontend/individual_pedigree.tsx
airenzp: ok

danr: Right one can have props that are not supposed to change. One way to enforce that they don't change is to remount the component using `key`: 
https://www.saltycrane.com/blog/2019/12/how-remount-react-component-when-prop-changes/
I think this is slightly weird since this is determined at the call site of the component rather than at the definition site of the component. It would make sense if one could indicate this when defining the component.

frontend/individual_pedigree.tsx
```diff
@@ -0,0 +1,100 @@
...
+      data => mounted && data && setPedigree(data),
+      error => console.error(error)
+    )
+
+    return () => { mounted = false }//allows to validate the mount state of the component
+
+  }, [id, generations_input])//dependent on these values
+
+  return <>
+    {individual && pedigree && <>//the component will be rendered if the individual and pedigree were retrieved
```
norling: this comment is rendered on the page :) it needs to be wrapped up as `{/* comment */}`

norling (approved):

frontend/individual_pedigree.tsx
airenzp: upss

airenzp: solved

PR #99, opened 2020-08-26
Feature/admin rework

norling:
Updates manage pages to more user friendly forms, as well as making database queries atomic (making sure that peewee doesn't lock on failed queries).
This PR does NOT include form validation as I figured that it was better to first review the forms themselves, and have a smaller PR (the same goes for field access restrictions).


frontend/userForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+  /**
+   * Converts a herd database id to a human readable label formatted as:
+   * '<herd-id>' or '<herd-id> - <herd-name>' depending on if the herd has a
+   * herd name or not.
+   *
+   * @param herdId database id of the herd to convert.
+   */
+  const herdIdToLabel = (herdId: number) => {
+    let allHerds: Herd[] = [];
+    allHerds = allHerds.concat.apply(allHerds, genebanks.map((g: Genebank) => g.herds));
```
danr: `allHerds` is `[]` so you can go for `const allHerds = genebanks.flatMap(g => g.herds)`

Also you don't need apply these days, use `...`: `[].concat(...xs, ...ys)`, but in this case you can just use array literals `[...xs, ...ys]`

frontend/userForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+      },
+      error => console.error(error)
+    );
+  }
+
+  /**
+   * Loads the user information for `id`, or sets the form to accept a new user.
+   */
+  React.useEffect(() => {
+    setUser({...defaultValues})
```
danr, line 95: You don't need to copy the object here before setting if you maintain the invariant to never mutate the user object you get from the state (which should be maintained)

frontend/userForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+  const [genebank, setGenebank] = React.useState(genebanks.length > 0 ? {value: genebanks[0].id, label: genebanks[0].name} : null)
+  const [herd, setHerd] = React.useState(null as any)
+  const history = useHistory()
+  const classes = useStyles();
+
+  const loadUser = (id: number) => {
+    get(`/api/manage/user/${id}`).then(
+      data => {
+        if (data) {
+          setUser(data)
```
danr: You can use ReactDOM.unstable_baitchedUpdates here around the two `set`s to not get two updates on the component. It's a gotcha to think about when handling callbacks from network requests https://itnext.io/react-setstate-usage-and-gotchas-ac10b4e03d60

frontend/userForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+   * Sends an UPDATE request to update a user role in the database. Reloads the
+   * user data on success.
+   * @param operation an object describing an update operation, formatted like:
+   *     {action: 'remove' | 'add',
+          role: 'owner' | 'manager' | 'specialist',
+          user: userId,
+          genebank: genebankId, if role is 'manager' or 'specialist'
+          herd: herdId, if role is 'owner'
+   */
+  const updateRole = (operation: any) => {
```
danr: Write the type information as a `type = ...` instead of in a comment! ;)

frontend/userForm.tsx
```diff
@@ -0,0 +1,306 @@
...
+    if (genebankId === undefined || genebank === null) {
+      return []
+    }
+
+    const currentGenebank = genebanks.find((g: Genebank) => g.id == genebank.value);
+    if (currentGenebank === undefined) {
+      return []
+    }
+    return currentGenebank.herds.map((h: Herd) => {
+      return {value: h.id, label: `${h.herd}${h.herd_name ? ` - ${h.herd_name}` : ''}`}
```
danr: Perhaps make a helper function of this label pattern (which is surisingly hard to read with a ternary nested in a template string)

frontend/herdForm.tsx
```diff
@@ -3,115 +3,288 @@
...
         data => {
-          data && setHerd(data);
-          setLoading(false);
+          if (data) {
+            // verify the data doesn't have nulls
+            Object.keys(data).forEach((k: string) => {data[k] = data[k] ?? ''})
+
+            // split physical address into address, postcode, postcity
+            if (data?.physical_address && data.physical_address.includes('|')) {
+              const addressParts = data.physical_address.split('|')
```
danr: Consider
```typescript
const  [address, postcode, postcity] = data.physical_address.split('|')
```

frontend/communication.tsx
```diff
@@ -62,3 +63,45 @@ export async function update(url: string, content: any) {
...
+
+/**
+ * Sends a request to create a new herd using the information in `herd`.
+ * Returns 'success' on success.
+ *
+ * @param herd Herd information.
+ */
+export async function createHerd(herd: Herd) {
+  // replace empty strings with null values
+  Object.keys(herd).forEach((k: string) => {herd[k] = herd[k] == '' ? null : herd[k]})
```
danr: Is there a potential problem there that you mutate your input?

frontend/herdForm.tsx
```diff
@@ -3,115 +3,288 @@
...
+    }
   }
 
   return <>
-    {loading && <h2>Loading...</h2> ||  herd &&
+    {loading && <h2>Loading...</h2> ||
       <>
-        <h1>{herd.name || `Besättning ${herd.id}`}</h1>
-        <form className={classes.form} noValidate autoComplete="off">
-          {simpleFields.map((field, i) => {
```
danr, line 81: simpleFields seemed like a reasonable abstraction here

danr:
In encam I run both pretty and typecheck and I find both helpful. Perhaps you want to too 
https://github.com/NBISweden/encam/blob/660e8d060b5225a0d7148e3f2e980d11a4988b9b/frontend/package.json#L34-L35

frontend/userForm.tsx
norling: Nice! Thanks! :)

frontend/userForm.tsx
norling: Yeah... It just felt better to make a copy than to potentially take the risk of mutating an object that should remain static!

frontend/userForm.tsx
norling: yeah - I can't really say why I just kept re-writing it instead of finally making a function :D

frontend/communication.tsx
norling: I figured it wasn't, but I can make a copy to be sure!

frontend/herdForm.tsx
norling: It did - but since the next step will be to start adding validation and (probably) permission levels I'll keep them separate until I know what the final form will look like. I agree that this is a very verbose form though!

jhagberg:
Looks like Herd list is broken and I cant get any herd information listing.

jhagberg:
Also, Its a bit confusing in the Besättningsinformation form the select list for individuals. What is that for? And are you supposed to change individuals there?

jhagberg:
It looks like you can't be "manager for both Gotlands and Mellerud" It's probably very unlikely. But should be possible or at least give an error

jhagberg:
Is it possible to only show "Besättning" when it's a valid input to minimize confusions

jhagberg (changes requested):
See comments for suggestions

frontend/data_context_global.tsx
```diff
@@ -62,6 +62,14 @@ export interface Herd {
     individuals?: Individual[]
 }
 
+export function herdLabel(herd: Herd): string {
```
danr, line 65: Elegant! :sunglasses:

danr (approved):

jhagberg (approved):
LGTM and working

pontus (approved):
I've only looked at database code but have no objections there (although most of the wrappings with atomic only address reading data, where it's superfluous, but I don't see any scenario where having transactions started for that would hurt).

PR #102, opened 2020-09-02
Use herd tracking to get herd for individuals

norling:
I think that this fixes all instances where `herd` was used erroneously. I have renamed `herd` as `origin_herd` to avoid this mistake in the future, so you will have to run `ALTER TABLE individual RENAME herd TO origin_herd;` in the database before testing this.

### DIFF
--- a/app/utils/data_access.py
+++ b/app/utils/data_access.py
@@ -15,6 +15,7 @@ from werkzeug.security import (
 from utils.database import (
     DB_PROXY as DATABASE,
     Herd,
+    HerdTracking,
     Individual,
     User,
 )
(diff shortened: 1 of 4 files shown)


andkaha (approved):
Can't say much about the Python bits, but the SQL looks correct.

app/utils/data_access.py
```diff
@@ -143,7 +144,22 @@ def get_herd(herd_id, user_uuid=None):
             data = herd.filtered_dict(user)
             if data["genebank"] not in user.accessible_genebanks:
                 return None
-            query = Individual().select().where(Individual.herd == herd)
+
+            # This turned out to be really hard to write in peewee
```
jhagberg, line 148: so what is the drawbacks with "raw"?

app/utils/data_access.py
```diff
@@ -143,7 +144,22 @@ def get_herd(herd_id, user_uuid=None):
             data = herd.filtered_dict(user)
             if data["genebank"] not in user.accessible_genebanks:
                 return None
-            query = Individual().select().where(Individual.herd == herd)
```
pontus, line 146: I think we should update the description for this function. Also, no dependencies on the old behaviour? We have no need for historical data through this function (no flag for all animals)?

jhagberg (approved):
LGTM

pontus (approved):

app/utils/data_access.py
pontus: I think this has been mentioned elsewhere, but it changes from working with whatever peewee supports to whatever peewee supports and our syntax is compatible with (likely postgresql only in the long run).

PR #103, opened 2020-09-02
Feature/messages

norling:
Adds a context for displaying user messages as snackbar pop-ups.

### DIFF
--- a/frontend/message_context.tsx
+++ b/frontend/message_context.tsx
@@ -14,7 +14,7 @@ export interface MessageContext {
 }
 const emptyContext: MessageContext = {
     userMessage() {},
-  }
+}
 
 export const MessageContext = React.createContext(emptyContext)
 
--- a/frontend/message_context.tsx
+++ b/frontend/message_context.tsx
@@ -1,0 +1,57 @@
+import * as React from 'react'
+import { Snackbar } from '@material-ui/core';
+import MuiAlert from '@material-ui/lab/Alert';
+
+/**
+ * The message context allows other complenents to display user messages in a
+ * consistent manner.
+ */
+
+export type MessageLevel = 'error' | 'warning' | 'info' | 'success';
+
+export interface MessageContext {
+    userMessage(msg: string, level: MessageLevel): void,
+}
+const emptyContext: MessageContext = {
+    userMessage() {},
+  }
+
+export const MessageContext = React.createContext(emptyContext)
+
+/**
+ * Exports the context variables and functions to be used by other components
+ */
+export function useMessageContext(): MessageContext {
+  return React.useContext(MessageContext)
+}
+
+export function WithMessageContext(props: {children: React.ReactNode}) {
+  const [severity, setSeverity] = React.useState('info' as MessageLevel)
+  const [message, setMessage] = React.useState('' as string)
+  const [open, setOpen] = React.useState(false);
+
+  function userMessage(message: string, severity: MessageLevel) {
+      setMessage(message);
+      setSeverity(severity);
+      setOpen(true);
+  }
+
+  const handleClose = (event: any, reason: string) => {
+      if (reason === 'clickaway') {
+          return;
+      }
+      setOpen(false);
+  }
+
+  return (
+    <MessageContext.Provider value={{userMessage}}>
+      {props.children}
+      <Snackbar open={open} autoHideDuration={5000} onClose={handleClose}>
+          <MuiAlert elevation={6} variant='filled' onClose={handleClose} severity={severity}>
+              {message}
+          </MuiAlert>
+      </Snackbar>
+    </MessageContext.Provider>
+  )
+}
+
--- a/frontend/userForm.tsx
+++ b/frontend/userForm.tsx
@@ -126,13 +126,14 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
             break; // updated user
           case "success":
           case "created":
+            const newUserId = data.data;
             unstable_batchedUpdates(() => {
               setNew(false);
-              users.push({email: user.email, id: data.data, name: user.email})
-              setUsers(Object.assign([], users))
+              const new_user = {email: user.email, id: data.data, name: user.email}
+              setUsers([...users, new_user])
               userMessage('User saved', 'success')
             })
-            history.push(`/manage/user/${data?.data}`)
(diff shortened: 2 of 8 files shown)


frontend/userForm.tsx
```diff
@@ -113,12 +115,17 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
...
         switch (data.status) {
-          case "updated": console.info("updated."); break; // updated user
+          case "updated":
+            userMessage('Changes saved', 'success')
+            break; // updated user
           case "success":
-            history.push(`/manage/user/${data.id}`);
+            history.push(`/manage/user/${data.id}`)
             setNew(false);
+            userMessage('User saved', 'success')
```
danr: Could wrap these in an unstable batched updates (since they are from a callback not controlled by react)

frontend/message_context.tsx
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+export type MessageLevel = 'error' | 'warning' | 'info' | 'success';
+
+export interface MessageContext {
+    userMessage(msg: string, level: MessageLevel): void,
+}
+const emptyContext: MessageContext = {
+    userMessage() {},
+  }
```
danr: Wonky indentation :sunglasses:

danr:
Looks good!!!

danr (approved):
Looks good!!! Now as a review approve!!!

frontend/message_context.tsx
jonandernovella: Perhaps a linter would be useful!

frontend/message_context.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [open, setOpen] = React.useState(false);
+
+  function userMessage(message: string, severity: MessageLevel) {
+      setMessage(message);
+      setSeverity(severity);
+      setOpen(true);
+  }
+
+  const handleClose = (event: any, reason: string) => {
+      if (reason === 'clickaway') {
```
jonandernovella, line 40: Enum would be better here I think.

frontend/userForm.tsx
```diff
@@ -113,12 +115,17 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
     protocol(`/api/manage/user/${id == 'new' ? 0 : id}`, postData).then(
       data => {
         switch (data.status) {
-          case "updated": console.info("updated."); break; // updated user
+          case "updated":
+            userMessage('Changes saved', 'success')
```
jonandernovella, line 125: The states such as "success" should be rewritten as enums/consts.

jonandernovella (changes requested):
A linter needs to be set up. Look at my comments also about enum and constants. Otherwise, really nice work!

frontend/userForm.tsx
norling: Yeah - I should make a plan for legal statuses and messages!

frontend/message_context.tsx
danr: No, a pretty-printer would!

frontend/userForm.tsx
danr: `data` is json which has no support for enums/consts, so strings is to be preferred

frontend/message_context.tsx
danr: `reason` is sent (as a string) from the ui library not created by us

frontend/userForm.tsx
jonandernovella: no need to send the enum object

danr: Ah right you meant the `success` message. There is no need to use an enum in typescript since there are singleton string types. Those are pretty sweet :v: All the possible message states are explicitly listed in the type declaration here:  https://github.com/NBISweden/theherdbook/pull/103/commits/2cf570b57107041515d1b3447eac9bab3837ca5d#diff-d9fdfb04a1ffc9d28baed006435248a5R10

frontend/userForm.tsx
```diff
@@ -126,11 +126,13 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
             break; // updated user
           case "success":
           case "created":
-            history.push(`/manage/user/${data?.data}`)
             unstable_batchedUpdates(() => {
               setNew(false);
+              users.push({email: user.email, id: data.data, name: user.email})
+              setUsers(Object.assign([], users))
```
danr: Never seen Object.assign on arrays :open_mouth: 

You can use `setUsers([...users, {email: user.email, id: data.data, name: user.email}])`. This way you don't need to mutate the array.

Maybe it's slightly clearer on two lines
```typescript
  const new_user = {email: user.email, id: data.data, name: user.email}
  setUsers([...users, new_user])
```

Is both name and email supposed to be user.email?

frontend/userForm.tsx
```diff
@@ -126,11 +126,13 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
...
               userMessage('User saved', 'success')
             })
+            history.push(`/manage/user/${data?.data}`)
```
danr: I would just introduce `const id = data?.data` after `case "created"` and just use this at both locations. Makes it a bit clearer.

frontend/userForm.tsx
norling: yeah, there're no usernames, so I use the e-mail for both. I could set `name` to `undefined` I guess... might be clearer... And yeah, your syntax is prettier! :D

frontend/userForm.tsx
norling: good point! definitely increases readability!

frontend/userForm.tsx
```diff
@@ -126,13 +126,14 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
             break; // updated user
           case "success":
           case "created":
+            const newUserId = data.data;
             unstable_batchedUpdates(() => {
               setNew(false);
-              users.push({email: user.email, id: data.data, name: user.email})
-              setUsers(Object.assign([], users))
+              const new_user = {email: user.email, id: data.data, name: user.email}
```
danr: use `id: newUserId` too :dancing_men:

norling: :D

jonandernovella (approved):

PR #105, opened 2020-09-07
Feature/new genebank view

norling:
This is mostly a removal of the old herd browsing pages in favor of just using the individuals table. That table is to be updated further, but as this part makes changes to the API I figured it's better to make an initial PR to make it easier for other people to keep working without having sync issues later.


airenzp (changes requested):
When building the pedigree the id should be the number of the individual, since this is the value used now to navigate through the individuals. See lines updated:
`
id = ind["number"]`
`if mother:
    add_parent(mother["number"])`
`if father:
    add_parent(father["number"])
`
 
And then: 
`parent = da.get_individual(parent_id, user_id)`

airenzp:
The interface is looking nicer :)

app/herdbook.py
```diff
@@ -214,7 +214,7 @@ def herd(h_id):
     data = da.get_herd(h_id, session.get("user_id", None))
     return jsonify(data)
 
-@APP.route("/api/individual/<int:i_id>")
+@APP.route("/api/individual/<i_id>")
 @login_required
```
airenzp, line 218: I have now removed the cached from the decorators in my branch since it is not considering different users.  Do the same here, to test the latest version of your code without the cache interfering.
`-@cache.cached(timeout=36000)`

airenzp (changes requested):
Good. I forgot to take a look at the herd pedigree. In order to make it work please remove also there the cache decorator and add this update in the individual short_info method:
`
 father = {"id": self.father.id, "number": self.father.number} if self.father else None`
` mother = {"id": self.mother.id, "number": self.mother.number} if self.mother else None
`

airenzp (approved):

jhagberg (commented):
LGTM

PR #109, opened 2020-09-11
Feature/improved table

norling:
basic functionality for the individuals table. Some minor quirks with sorting, but I don't want to spend the time fixing it until I've fixed the page-slowdown bug (since I might need to change the material-table to some other table if I can't fix it).


jhagberg (approved):
LGTM, Mostly focused on functions

frontend/genebanks.tsx
```diff
@@ -1,135 +1,66 @@
...
+    } else if (genebanks.length > 0) {
+      const targetGenebank = genebanks.find((g: Genebank) => g.name.toLowerCase() == subpath.toLowerCase())
+      if (targetGenebank) {
+        setGenebank(targetGenebank)
+      }
     }
-    setGenebank(targetGenebank);
-    setIndividuals(targetGenebank.individuals);
-  }
+  }, [subpath, genebanks]);
```
danr, line 44: Some room for improvement here:
* `subpath` does not need to be in a React state
* no need to set the genebank again if it is already the correct one
```typescript 
  const subpath = location.pathname.replace(url, '').trim().replace(/\//, '')

  React.useLayoutEffect(() => {
    if (!subpath && genebanks.length > 0) {
      history.push(`${url}/${genebanks[0].name}`)
    } else if (genebanks.length > 0) {
      const targetGenebank = genebanks.find((g: Genebank) => g.name.toLowerCase() == subpath.toLowerCase())
      if (targetGenebank && targetGenebank !== genebank) {
        setGenebank(targetGenebank)       
      }
    }
  }, [subpath, genebank, genebanks]);
```

frontend/genebanks.tsx
```diff
@@ -1,135 +1,66 @@
...
+                        value={g.id}
+                        color={genebank && genebank.id == g.id ? 'primary' : 'default'}
+                        >
+                  {g.name}
+                </Button>
+              </Link>
+          })
+        }
       </div>
+      { genebank && <GenebankView genebank={genebank} /> }
```
danr: Right now `GenebankView` can be unnecessarily redrawn with the same genebank. This can be fixed by either wrapping the `GenebankView` component in `React.memo` or using `React.useMemo` here:
```typescript
      { React.useMemo(() => genebank && <GenebankView genebank={genebank} />, [genebank]) }
```

frontend/genebanks.tsx
```diff
@@ -1,135 +1,66 @@
...
-  const selectGenebank = (id: number) => {
-    if (!genebanks) {
-      return;
-    }
-    const targetGenebank = genebanks.find((g: Genebank) => g.id == id);
-    if (!targetGenebank) {
-      return;
+  React.useEffect(() => {
+    if (!subpath && genebanks.length > 0) {
+      history.push(`${url}/${genebanks[0].name}`)
```
danr, line 36: Ideally this component should not mess around with the location. It would be nice if this was handled somewhere else and the component just got the relevant url pierce as a prop. I don't have a suggestion though so we can skip it for now.

danr (changes requested):
Found some things in Genebanks component that can be fixed

frontend/genebanks.tsx
norling: Why? doesn't it make sense to have the page relocate to make sure that the actions are available in history?

frontend/genebanks.tsx
norling: nice! I'll use that! :)

frontend/genebanks.tsx
norling: yeah, that's prettier :) For some reason I got into my head that use*Effect could only monitor state-variables... dunno why ^^ Good to check so that the same genebank isn't reset as well!

frontend/genebanks.tsx
danr: I think browstery history is a great feature, I just wasn't sure it should be in this component. But thinking about it now I realize that this is the navigation component for genebanks, so this should be about right.

danr (approved):

PR #110, opened 2020-09-11
Feature/snowpack

danr:
Change dev server to snowpack, building with esbuild.

* Snowpack dev server with a proxy to the nginx backend. This is a clear improvement over parcel since it does not have a gigantic memory leak + hot module reloads are around 100ms instead of >1000ms.
* Building with esbuild. This is because snowpack creates an incorrect bundle. (It seems to remove too much from @material-ui/core.)

Esbuild does not support css imports (yet) so there is a workaround in pedigree.tsx.


danr:
With `docker compose up --build` the dev server is on docker host localhost:8080 and nginx on localhost:4200.

app/herdbook.py
```diff
@@ -266,7 +266,8 @@ def build_herd_pedigree(id):
     return result
 
 
-@APP.route('/api/pedigree/<i_id>', defaults={"generations": 5})
+# Had to remove the default value of generations because it causes some kind of redirect
+# (308 moved permanently?) messing up the snowpack proxy.
 @APP.route('/api/pedigree/<i_id>/<int:generations>')
 @login_required
 def pedigree(i_id, generations):
```
norling: You can add the default value here as `generations=5`.

pontus:
As this is still under active development, I'd strongly prefer keeping the development flow as close as possible to the intended production flow. In this case, that means I'd prefer keeping the nginx setup for development to avoid surprises down the line. 

(It's, of course, possible that there are benefits of using only snowpack that outweigh those concerns, but it's not obvious to me. This is also somewhat coupled to our current lack of robust front end/round trip testing.)

jhagberg:
> As this is still under active development, I'd strongly prefer keeping the development flow as close as possible to the intended production flow. In this case, that means I'd prefer keeping the nginx setup for development to avoid surprises down the line.
> 
> (It's, of course, possible that there are benefits of using only snowpack that outweigh those concerns, but it's not obvious to me. This is also somewhat coupled to our current lack of robust front end/round trip testing.)

What changes do you suggest?

airenzp:
The PR is looking overall good to me. Nice cleanup. I still need though to move the node_modules though outside of the mounted frontend volume.  There is a mac bug in docker, I think,  that overwrites the volume during runtime. To solve it I add a .yarnrc file in my local branch and create a folder in the container for the node_modules. However, after moving the node_modules I can see the rebuild happening but I get no updates on the interface. I will continue to prove it today

danr:
@pontus So do you prefer nginx to forward to the snowpack dev server, similar to how it forwarded to the parcel dev server before?

danr:
@airenzp Sounds good, can you push your mac fixes to a new branch from this one so that I can try it out?

pontus:
Yes, that would be my clear preference in this case (others may have other opinions).

norling (approved):
Seems to work well, and is faster than parcel! :) Haven't done extensive testing, so I'll just assume that you have!

airenzp (approved):

app/herdbook.py
danr: Should be fixed now

airenzp:
@danr I am still trying to solve the compilation with snowpack as well, after rebuilding from scratch not even using the --node-modules parameter with a path to a folder outside of the mounted volume worked. The build command fails. You do not need to waste your time with it, but if you want to try it, this is the change: https://github.com/NBISweden/theherdbook/compare/feature/snowpack...feature/snowpack-mac?expand=1

airenzp:
Hej, one thing to note. The port 8080 is not reflecting the changes in the code, for development 4242 needs to be used, this was making me crazy. It actually works for MAC as well!!! This should be clarified in the README.md

PR #111, opened 2020-09-14
Feature/loading scripts

andkaha:
New loading scripts added.

These scripts load the CSV data into a `data` table and uses SQL to populate the tables of the main database from there.  A database dump file of the loaded tables is then created.

See also the `README` file.

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -97,30 +97,32 @@ END_SQL
 # Load tracking data for years 2000 through to 2020
 year=2000
 while [ "$year" -le 2020 ]; do
+	column=$year
+
 	cat <<-END_SQL
 		-- Fix column type etc.
-		ALTER TABLE data ALTER "$year" TYPE VARCHAR(10);
-		UPDATE data SET "$year" = CONCAT('G', "$year")
-		WHERE "$year" IS NOT NULL AND "$year" NOT LIKE 'G%';
+		ALTER TABLE data ALTER "$column" TYPE VARCHAR(10);
+		UPDATE data SET "$column" = CONCAT('G', "$column")
+		WHERE "$column" IS NOT NULL AND "$column" NOT LIKE 'G%';
 
 		-- Add missing herds
 		INSERT INTO herd (genebank_id, herd)
-		SELECT	DISTINCT gb.genebank_id, d."$year"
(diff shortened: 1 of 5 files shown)


jhagberg (changes requested):
We get the data as XLS but this takes CSV can we incorporate so we can load the original file directly instead of using some external thing to translate the xls file to CSV?

jhagberg:
in2csv is a command in csvkit can that be used instead of the Google step?

andkaha:
I will investigate the `in2csv` utility (which I haven't tried using before).

scripts/load-gotland.sh
```diff
@@ -0,0 +1,170 @@
...
+done | psql
+
+# The Gotland data set has herd names etc. in a separate CSV file.
+# Load that file separately (using static filename here for now).
+
+csvsql	--db "postgresql://$PGUSER:dummy@/$PGDATABASE" \
+	--tables data2 \
+	--overwrite \
+	--skip-lines 5 \
+	--insert herd-registry-gotland.csv
```
jhagberg: There is no information in the Readme about this file

jhagberg (changes requested):

scripts/load-gotland.sh
andkaha: Will fix that later in the afternoon!

andkaha: This is now addressed in `README`, I hope.

jhagberg: Using in2csv --skip-lines 5 worked for that file xlsx file I think! Should we specify the name on the cmd line also?

jhagberg:
When running the script I get "expected to find ../app/config.ini" so I guess you need to build the docker images first. I did this but probably in another branch

scripts/load-gotland.sh
andkaha: This (using `--skip-lines 5`) is what I did in the update version.

Since the auxiliary file is not expected to change very often, I opted for hard-coding its name in the script.  This is in contrast to the other data files. These may change often.

Also, I kinda hope these loading script should be of _temorary_ use until such time that data entry happens via the web site.

andkaha:
I merged this as all review issues have been addressed, and since no further serious issues have surfaced.  Further changes to the loading script code will be addressed in future PRs.

PR #112, opened 2020-09-18
Feature/username

norling:
Adds an optional username to all users. To use this you need to add the username column to the user table in the database.

connect using: `docker exec -ti herdbook-db psql -U herdbook herdbook`
then run: `ALTER TABLE hbuser ADD COLUMN IF NOT EXISTS username TEXT UNIQUE;`

This allows admins and manager to give users usernames, and users to log in with either username or password.

### DIFF
--- a/app/utils/data_access.py
+++ b/app/utils/data_access.py
@@ -71,22 +71,23 @@ def register_user(email, password, validated=False, privileges=[]):
         user.save()
     return user
 
-def authenticate_user(email, password):
+def authenticate_user(name, password):
     """
-    Authenticates an email/password pair against the database. Returns the
-    user info for the authenticated user on success, or None on failure.
+    Authenticates an email or username and password against the database.
+    Returns the user info for the authenticated user on success, or None on
+    failure.
     """
     try:
         with DATABASE.atomic():
-            user_info = User.get(User.email == email)
+            user_info = User.select().where((User.email == name) | (User.username == name)).get()
         if check_password_hash(user_info.password_hash, password):
-            logging.info("Login from %s", email)
+            logging.info("Login from %s", name)
             return user_info
     except DoesNotExist:
         # Perform password check regardless of username to prevent timing
         # attacks
         check_password_hash("This-always-fails", password)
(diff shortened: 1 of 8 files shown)


jhagberg:
Starting from scratch do you still need to do the manual step of altering table?

norling:
> Starting from scratch do you still need to do the manual step of altering table?

No - the database is created as it should be if it doesn't exist :)

jhagberg (approved):
LGTM

jhagberg:
> > Starting from scratch do you still need to do the manual step of altering table?
> 
> No - the database is created as it should be if it doesn't exist :)

Yhea I missed the changes in the file database.py good!

frontend/src/userForm.tsx
```diff
@@ -227,6 +233,12 @@ export function UserForm({id}: {id: number | 'new' | undefined}) {
                      value={user.email ?? undefined}
                      onChange={e => setUser({...user, email: e.target.value})}
             />
+            <TextField label='Användarnamn'
+                       type='username'
+                       className={classes.simpleField}
+                       value={user.username ?? undefined}
```
danr: Here `undefined` should be `''`. Otherwise the component can change between being controlled and uncontrolled.

app/utils/data_access.py
```diff
@@ -64,29 +65,33 @@ def register_user(email, password, validated=False, privileges=[]):
...
+    Authenticates an email or username and password against the database.
+    Returns the user info for the authenticated user on success, or None on
+    failure.
     """
+    if not name or not password:
+        return None
     try:
         with DATABASE.atomic():
-            user_info = User.get(User.email == email)
+            user_info = User.select().where((User.email == name) | (User.username == name)).get()
```
danr, line 86: Can we get a problem if one user's username is another user's email address?

danr (approved):
Looks jolly!

app/utils/data_access.py
norling: yes - that will be solved later when I apply validation and rules, which will include emails needing an '@', and usernames disallowing special characters though.

PR #113, opened 2020-09-24
Feature/rerender fixes

danr:
Ok after spending quite some time with the React profiler I figured out that there were just a few sources of inefficiencies:

* The `Routed` helper recreated an `Inner` function component making reconciliation fail every time.
* Navigation looks at the user context so we want to keep the changes to it as few as possible. The user context is now memoized and depends on loadData from the data context so this function is memoized too.

I moved the PedigreeCSS to an own file too, this is just a bonus for this PR. :thinking:


danr:
The remaining time sink is the load of one of the genebanks which is at 10MB.

airenzp (approved):
Nice cleanup. Great that the contexts are not affecting the rendering anymore.

PR #116, opened 2020-09-29
Proxy from nginx to snowpack dev server only

danr:
Changing to this setup by @pontus from before because the snowpack dev server doesn't proxy `UPDATE` properly.

Dev server is now at localhost:8443 and localhost:8080 again.

### DIFF
--- a/system/herdbook.devel.nginx
+++ b/system/herdbook.devel.nginx
@@ -1,0 +1,22 @@
+    server {
+        listen 8443 ssl http2;
+        listen 8080;
+
+        ssl_certificate /code/cert.pem;
+        ssl_certificate_key /code/key.pem;
+        root /app/dist/;
+
+        location /api/ {
+                # There were issues using the wsgi transport for uwsgi,
+                # use its built in http server.
+                proxy_pass http://127.0.0.1:9090/api/;
+        }
+
+        location / {
+               proxy_pass http://herdbook-frontend-devel:4242;
+               proxy_http_version 1.1;
+               proxy_set_header Upgrade $http_upgrade;
+               proxy_set_header Connection "Upgrade";
+               proxy_set_header Host $host;
+        }
+    }
--- a/.docker/main
+++ b/.docker/main
@@ -13,6 +13,7 @@ RUN cd /app/dist && ln -s . static
 
 COPY system/supervisord.conf /etc/
 COPY system/nginx.conf /etc/nginx/
+COPY system/herdbook.devel.nginx /etc/nginx/herdbook.devel
 COPY system/herdbook.standalone.nginx /etc/nginx/herdbook.standalone
 
 COPY system/entrypoint.sh /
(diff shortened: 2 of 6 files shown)


system/herdbook.devel.nginx
```diff
@@ -0,0 +1,22 @@
...
+        location /api/ {
+                # There were issues using the wsgi transport for uwsgi,
+                # use its built in http server.
+                proxy_pass http://127.0.0.1:9090/api/;
+        }
+
+        location / {
+               proxy_pass http://herdbook-frontend-devel:4242;
+               proxy_http_version 1.1;
+               proxy_set_header Upgrade $http_upgrade;
```
danr, line 18: Do we need these three `proxy_set_header`? @pontus

pontus: Does snowpack use a wss to transport state information back and forth? In that case yes (at least the Upgrade ones).

pontus (commented):
If onw wants to serve static content first properly, we probably need to do something along the lines of https://github.com/NBISweden/theherdbook/blob/766ecdcfcc499845a4259bd286632148c40da0ff/system/herdbook.devel.nginx, alternatively serve at least .well-known statically (or decide that isn't an important use case for the devel, mostly depends on whatever one wants to be able to get ACME certs [e.g. letsencrypt] through what's offered, can likely be achieved by pointing  to a statically served area within what snowpack sees or using another method)
.

pontus (approved):

danr:
> Does snowpack use a wss to transport state information back and forth? In that case yes (at least the Upgrade ones).

OK! Yes, it uses a websocket. (It only pushes information to the browser though)

> If onw wants to serve static content first properly, we probably need to do something along the lines of https://github.com/NBISweden/theherdbook/blob/766ecdcfcc499845a4259bd286632148c40da0ff/system/herdbook.devel.nginx, alternatively serve at least .well-known statically (or decide that isn't an important use case for the devel, mostly depends on whatever one wants to be able to get ACME certs [e.g. letsencrypt] through what's offered, can likely be achieved by pointing to a statically served area within what snowpack sees or using another method)

I don't understand at all. What part of the linked file are you refrring to? Then what is _serve well-known statically_? Also what does the rest of the paragraph mean?

pontus:
Sorry, lines 23-25 (to replace 15-21 in https://github.com/NBISweden/theherdbook/blob/b558e7ec567c140665332882380059db4cf7e2f5/system/herdbook.devel.nginx, but with the header setting added) and 33-36.

And the use case that is supported by that is serving the same static content in development as is served in production (config/updates is brought in).

And one use case for that is getting certificates through lets encrypt with the "host a file" method, which means your client puts a file that is served as `.well-known/something`.

norling (approved):
It Works™!

danr:
Now I understand much better, thanks @pontus. Is it important to run let's encrypt while developing though? Or would that be to test certbot setup on the dev server?

pontus:
Probably not an important use case. My original impression when designing was to support frequent refreshes (git pull) of a user visible site.

In practice, I think PRs have been rather large and frequently done changes requiring full rebuilds anyway, so having the user visible site "static" is not that big of a drawback.

danr:
OK let's not bother with that until it becomes important then imo

PR #119, opened 2020-10-05
Feature/frontend pedigree

norling:
Moves pedigree calculations from backend to frontend.

There are a bunch of things that can be optimized in this, in particular, an index of all individuals would probably speed this up by a factor of at least 10. I figured this might work well enough for now though, and we can revisit it once we've decided on a final way of displaying the pedigrees.


frontend/src/data_context_global.tsx
```diff
@@ -6,22 +6,35 @@ export interface NameID {
...
 }
 
+export interface LimitedIndividual {
+    id: number,
+    name: string | null,
+    number: string,
+    sex?: string,
+}
+
 export interface Individual {
```
danr: You can do `export interface Individual extends LimitedIndividual` to avoid repeating fields (however seems like the type of `sex` is slightly incompatible)

frontend/src/herd_pedigree.tsx
```diff
@@ -29,12 +31,12 @@ export function HerdPedigree({ id }: { id: string }) {
   ]
 
   React.useEffect(() => {
-    let mounted = true; // Indicate the mount state
+    const pedigree = herdPedigree(genebanks, id, 5);
+    setPedigree(pedigree)
+  }, [genebanks, id])
```
danr: I think you can instead of putting pedigree in React.useState I think you can do 
```typescript
const pedigree = React.useMemo(() => herdPedigree(genebanks, id, 5), [genebanks, id])
```

frontend/src/pedigree.tsx
```diff
@@ -1,73 +1,165 @@
...
-    },
+      return {nodes: pedigreeNodes, edges: pedigreeEdges}
+    }
+    const pedigree = getPedigree(indData)
+    nodes = [...nodes, ...pedigree.nodes]
+    edges = [...edges, ...pedigree.edges]
+  }
+  // remove duplicate nodes and edges
+  nodes = nodes.filter((v,i,s) => s.findIndex(o => o.id == v.id) == i)
+  edges = edges.filter((v,i,s) => s.findIndex(o => o.id == v.id) == i)
```
danr: Creative solution! I have this in encam: 

```typescript
/** Returns a copy of the array with duplicates removed, via reference equality

  uniq([8,3,8,1]) // => [8,3,1]

  const x = {}
  const y = {}
  uniq([x,x,y,x,y]) // => [x,y]

*/
export function uniq<A>(xs: A[]): A[] {
  const seen = new Set()
  return xs.filter(x => {
    const duplicate = seen.has(x)
    seen.add(x)
    return !duplicate
  })
}
```

You can use it if you want, it's complexity is better (not quadratic). You might want to add a `by`/`key` argument so you can write `nodes = uniq(nodes, v => v.id)`

frontend/src/pedigree_plot.tsx
```diff
@@ -0,0 +1,73 @@
...
+          const nodeid = params.nodes[0];
+          history.push("/individual/" + nodeid);
+        }
+      });
+    },
+    [pedigree]);
+
+
+  return <>
+    <PedigreeCSS />
```
danr, line 67: While you're working on these files, if you bump the esbuild version you can just import the pedigree css file directly!

danr (approved):
Looks good!~

frontend/src/pedigree_plot.tsx
danr: Oh, you'd need to add a link to the built `main.css` to the build.sh script too

airenzp (approved):
Looking good to me.

frontend/src/data_context_global.tsx
norling: good point! Once all the types are in place we should really go through them and see what we can smash together...

frontend/src/herd_pedigree.tsx
norling: yes! way neater!

frontend/src/pedigree_plot.tsx
norling: neat! :)

frontend/src/pedigree.tsx
norling: hmm. yeah that's probably for the best :) I tend to write these chain-expressions while designing, but I need to learn to go back and clean them up! :D

PR #120, opened 2020-10-07
R api with endpoints

jhagberg:
Docker file and docker-compose updates
Three endpoints 

/kinship/{genebank_id} 
returns the kinship matrix of genebank 1 or 2

​/inbreeding​/{genebank_id}
Returning a list of all inbreeding values for the selected genebank

/meankinship/{genebank_id}
Returns a list of the mean kinship (this could be calculated in python also from the kinship matrix

Swagger to test api is found here http://localhost:31113/__docs__/

### DIFF
--- a/.docker/r-api
+++ b/.docker/r-api
@@ -1,0 +1,21 @@
+FROM rocker/r-ver:4.0.2
+
+WORKDIR /usr/local/optisel
+
+RUN apt update \
+  && apt install -y --no-install-recommends \
+  xvfb \
+  libpq-dev \
+  libx11-dev \
+  pkg-config \
+  libglu1-mesa-dev \
+  zlib1g-dev \
+  libxml2-dev \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY . .
+
+RUN MAKEFLAGS='-j4' Rscript -e 'install.packages(c("here","RPostgres","optiSel","plumber","jsonlite","devtools"),clean=TRUE, Ncpus=8);\
+                    library(devtools);devtools::install_github("rmsharp/nprcgenekeepr")'
+
+ENTRYPOINT ["Rscript", "/usr/local/optisel/R/serve.R"]
--- a/R/api.R
+++ b/R/api.R
@@ -1,0 +1,105 @@
+library(optiSel)
+library(nprcgenekeepr)
+library(DBI)
+library(jsonlite)
+# Returns a connection instance to a PG server
+connect_to_pg <- function() {
+  pg_host <- Sys.getenv("POSTGRES_HOST", "localhost")
+  con <- dbConnect(RPostgres::Postgres(),dbname = 'herdbook',
+                   host = get("pg_host"),
+                   port = 5432,
+                   user = 'herdbook',
+                   password = 'insecure')
+  return (con)
+}
+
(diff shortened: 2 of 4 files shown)


R/api.R
```diff
@@ -0,0 +1,105 @@
...
+# Returns a connection instance to a PG server
+connect_to_pg <- function() {
+  pg_host <- Sys.getenv("POSTGRES_HOST", "localhost")
+  con <- dbConnect(RPostgres::Postgres(),dbname = 'herdbook',
+                   host = get("pg_host"),
+                   port = 5432,
+                   user = 'herdbook',
+                   password = 'insecure')
+  return (con)
+}
```
norling, line 14: You should probably make R read the same settings file that python does, otherwise this will break once the database is secured.

docker-compose.yml
```diff
@@ -46,3 +45,17 @@ services:
...
+             dockerfile: .docker/r-api
+         environment:
+            - OPTISEL_PORT=31113
+            - OPTISEL_BIND_ADDR=0.0.0.0
+            - POSTGRES_HOST=herdbook-db
+         volumes:
+            - ./R:/usr/local/optisel/R
+         image: r-api
+         ports:
+             - 31113:31113
```
norling, line 61: It's fine to have the port open for debugging, but you don't need to as the containers can connect to each other anyway!

norling (changes requested):
I'd prefer to include the python connection in this PR to make the API usable from the main application, but if we prefer to take things in steps this is fine! :) 
You should sort out reading the settings file though, as that's pretty important.

R/api.R
pontus: Agreed, and the best way of doing is so probably to put these variables in a environment file and add  
```
        env_file:
            - .envfile
```

to all the services.

.docker/r-api
```diff
@@ -0,0 +1,21 @@
...
+  pkg-config \
+  libglu1-mesa-dev \
+  zlib1g-dev \
+  libxml2-dev \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY . .
+
+RUN MAKEFLAGS='-j4' Rscript -e 'install.packages(c("here","RPostgres","optiSel","plumber","jsonlite","devtools"),clean=TRUE, Ncpus=8);\
+                    library(devtools);devtools::install_github("rmsharp/nprcgenekeepr")'
```
pontus, line 19: Do we want to lock these to a specific version or is that not necessary?

pontus (approved):
Seems to work, format seems reasonable for parsing as well.

Timing is 4.5s for kinship, 4.0s for mean kinship and 2.6s for inbreedings on my machine.

.docker/r-api
jhagberg: Yes perhaps that is a good Idea at least R version can completely destroy this so that should be fixed version

docker-compose.yml
jhagberg: Yes copy paste from my side it should not be open!

R/api.R
jhagberg: Okey that sounds like a separate PR. Can you make a issue about that? But I agree

norling (approved):

PR #122, opened 2020-10-09
Change to use env files 

jhagberg:
Added env file in docker services.

Update R and python files.


app/utils/settings.py
```diff
@@ -4,30 +4,16 @@
...
 
-postgres.name = PARSER.get("postgres", "name")
-postgres.host = PARSER.get("postgres", "host")
-postgres.port = PARSER.get("postgres", "port")
-postgres.user = PARSER.get("postgres", "user")
-postgres.password = PARSER.get("postgres", "password")
+#Read configuration from environment variable
+
+postgres.name = os.environ.get("POSTGRES_DB", "herdbook")
+postgres.host = os.environ.get("POSTGRES_host", "herdbook-db")
```
jonandernovella: Should be `postgres.host = os.environ.get("POSTGRES_HOST", "herdbook-db")`

jonandernovella (changes requested):

app/utils/settings.py
jhagberg: oh yes

pontus (approved):

jonandernovella (approved):
LGTM

norling (changes requested):
This PR breaks our coding guidelines against not having sensitive data in repos. I'd prefer to add `database-variables.env` (and possibly `r-api-variables.env`) to `.gitignore` and allow the system to run on default values - or have a startup-script that creates the environment files with a secure password if the files doesn't exist.
The `README` should also be updated with information on how to change settings in these files, and `init_db.sh` should either be removed or updated as this PR will break that script (I'd be fine with removing it).

andkaha:
> This PR breaks our coding guidelines against not having sensitive data in repos. I'd prefer to add `database-variables.env` (and possibly `r-api-variables.env`) to `.gitignore` and allow the system to run on default values - or have a startup-script that creates the environment files with a secure password if the files doesn't exist.
> The `README` should also be updated with information on how to change settings in these files, and `init_db.sh` should either be removed or updated as this PR will break that script (I'd be fine with removing it).

Using a `.default` file would be in order, IMHO, just like we previously had with `app/config.ini.default`.

norling:
Yes! that would be fine! :) Are you still using `init_db.sh` btw?

andkaha:
> Yes! that would be fine! :) Are you still using `init_db.sh` btw?


Yes, so that script would possibly need to be updated(?)

It should at least contain an updated path to the correct "env file".

jhagberg:
> The `README` should also be updated with information on how to change settings in these files, and `init_db.sh` should either be removed or updated as this PR will break that script (I'd be fine with removing it).

But there is no information now on how to do this in the README so should it be an instruction on how to do it by hand or a init script?

README.md
```diff
@@ -44,6 +53,8 @@ f62cebbc2e43        herdbook_main              "/bin/sh -c /entrypo…"   3 hour
 
 ```
 
+If you 
+
```
norling: Some collaborators who are new to the project may have trouble understanding the nuances of this paragraph.

norling (approved):

PR #125, opened 2020-10-12
Feature/multi stage R build

jhagberg:
Speed up changes in build R docker if we change R code.


pontus (approved):
I believe this will work as intended, there is however still a volume in the `docker-compose.yml` to bind R at /usr/local/optisel/R that I expect can go away now.

PR #126, opened 2020-10-13
Feature/owner page

norling:
This PR was starting to grow a bit too large, so I think it's better to PR here than when it's feature complete.
Most of the changes are code rearrangement though!

### DIFF
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -74,6 +74,14 @@ export interface Herd {
     individuals?: Individual[]
 }
 
+export function individualLabel(individual: Individual | LimitedIndividual): string {
+    let label = `${individual.number}`
+    if (individual.name){
+        label += ` - ${individual.name}`
+    }
+    return label
+}
+
 export function herdLabel(herd: Herd): string {
     let label = `${herd.herd}`
     if (herd.herd_name) {
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -28,18 +28,18 @@ export interface LimitedIndividual {
 export interface Individual extends LimitedIndividual{
     herd: HerdNameID
     certificate: string | null
-    birthDate: string | null
+    birth_date: string | null
     mother: LimitedIndividual | null
     father: LimitedIndividual | null
-    color: string | null
-    colorNote: string | null
-    deathDate: string | null
-    deathNote: string | null
+    colour: string | null
+    colour_note: string | null
+    death_date: string | null
+    death_note: string | null
     litter: number | null
     notes: string | null
     weights: Array<DateValue> | null
     bodyfat: Array<DateValue> | null
-    herdTracking: Array<DateValue> | null
+    herd_tracking: Array<DateValue> | null
     herd_active: boolean
     active: boolean
     alive: boolean
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -74,7 +74,7 @@ export interface Herd {
     individuals?: Individual[]
 }
 
-export function individualLabel(individual: Individual | LimitedIndividual): string {
+export function individualLabel(individual: LimitedIndividual): string {
     let label = `${individual.number}`
     if (individual.name){
         label += ` - ${individual.name}`
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -154,8 +154,8 @@ export function FilterTable({individuals, title = '', filters = []}:
       return []
     }
     return individuals.filter((i: Individual) => {
-      for (let f = 0; f < currentFilters.length; f++) {
-        if (currentFilters[f].active && !i[currentFilters[f].field]) {
+      for (let filter of currentFilters) {
+        if (filter.active && !i[filter.field]) {
           return false;
         }
       }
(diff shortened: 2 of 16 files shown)


frontend/src/filter_table.tsx
```diff
@@ -0,0 +1,218 @@
...
+    })
+    setColumns(newColumns)
+  }
+
+  const filteredIndividuals = () => {
+    if (!individuals) {
+      return []
+    }
+    return individuals.filter((i: Individual) => {
+      for (let f = 0; f < currentFilters.length; f++) {
```
danr: `for (let f of currentFilters)` if you want to type less :nerd_face:

frontend/src/filter_table.tsx
```diff
@@ -0,0 +1,218 @@
...
+      { individuals
+        ? <>
+            {currentFilters.map(filter =>
+              <FormControlLabel key={filter.field}
+                control={<Checkbox
+                          name={filter.field}
+                          checked={filter.active}
+                          onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
+                            filter.active = e.target.checked
+                            setFilters([...currentFilters])
```
danr, line 193: `currentFilters` is already the filter! This is the same as `setFilters(currentFilters)`, which is the same as not setting! 

If you want to get around the mutation you can do something like 
```typescript
const next = [...currentFilters]
next[i] = {...next[i], active: e.target.checked } // i index in the map
setFilters(next)
```
:man_shrugging:

frontend/src/herd_view.tsx
```diff
@@ -4,55 +4,147 @@
...
+    cursor: 'pointer',
+    '&:hover': {
+      color: 'blue',
+    }
+  },
+});
+
+interface TabPanelProps {
+  children?: React.ReactNode;
+  dir?: string;
```
danr: `dir` doesn't seem to be used

frontend/src/herd_view.tsx
```diff
@@ -4,55 +4,147 @@
...
+  },
+  animalList: {
+    cursor: 'pointer',
+    '&:hover': {
+      color: 'blue',
+    }
+  },
+});
+
+interface TabPanelProps {
```
danr: Maybe you want `interface TabPanelProps extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement> {` to support `...other` below. Haha I wish this type was not so unwieldy.

frontend/src/individual_view.tsx
```diff
@@ -0,0 +1,215 @@
...
+        console.error(error);
+        userMessage(error, 'error')
+      }
+    )
+  }, [id])
+
+  /**
+   * Set children
+   */
+  React.useEffect(() => {
```
danr: You can use the `const children = React.memo( ... )` here to avoid useState + useEffect

frontend/src/owner.tsx
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Loads herd information from the backend and displays it as editable fields.
+ */
+export function Owner() {
+  const { user } = useUserContext()
+  const [ activeHerd, setActiveHerd ] = useState(undefined as string | undefined)
+  const style  = useStyles()
+
+  React.useEffect(() => {
+    setActiveHerd(user?.is_owner && user.is_owner.length > 0
```
danr: This can be the initial state for activeHerd I think

frontend/src/data_context_global.tsx
```diff
@@ -74,6 +74,14 @@ export interface Herd {
     individuals?: Individual[]
 }
 
+export function individualLabel(individual: Individual | LimitedIndividual): string {
```
danr: `individual: LimitedIndividual` is enough here, you can still call it with an `Individual`

danr (approved):
Lots of good stuff here! Comments are just nitpicks really. I can kind of see how React gets in your way...

frontend/src/filter_table.tsx
norling: Good point :D I don't know why I always forget this syntax... I mean it's odd, but I'm supposed to know it xD

frontend/src/filter_table.tsx
norling: I think it works due to the spread operator! `a = []; a != [...a]`.

frontend/src/herd_view.tsx
norling: ah - must be some remnant of a bygone idea! I'll remove it.

frontend/src/herd_view.tsx
norling: good point, and yeah, they could have named it `React.ElementProps` imho xD

frontend/src/individual_view.tsx
norling: good point! I'm still not quite used to `memo`!

frontend/src/owner.tsx
norling: yes!

frontend/src/data_context_global.tsx
norling: good point!

frontend/src/filter_table.tsx
danr: I think the memory rule is 
- What would be sensible?
- `for ... in ...`
- Then what does JavaScript use?
- `for ... of ...`

PR #149, opened 2020-10-19
Feature/frontend login check

norling:
Redirects restricted pages to /login when user isn't logged in.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -36,7 +36,7 @@
     "@material-ui/core": "^4.11.0",
     "@material-ui/icons": "^4.9.1",
     "@material-ui/lab": "^4.0.0-alpha.50",
-    "material-table": "1.67.0",
+    "material-table": "1.68.0",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-icons": "^3.11.0",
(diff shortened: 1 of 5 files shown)


danr:
How about making a `Restricted` component like this?

```typescript
function Restricted(props: {children: React.ReactElement}) {
  const {user} = useUserContext()
  const {location} = useHistory() // cannot remember the hook name
  return user ? children : <Redirect to={{ pathname: "/login", state: { from: location } }}/> 
}
```

Then you can keep using the same functions as before (and you don't need to break apart useRoutedTabs):

So instead of

```typescript
        <AuthRoute path="/herd-pedigree/:id">
￼          {params => <HerdPedigree id={params.id}/>}
￼       </AuthRoute>
```

you can do

```typescript 
        <ui.Routed path="/herd-pedigree/:id">
￼          {params => 
              <Restricted> 
                <HerdPedigree id={params.id}/>
              </Restricted>
           }
￼       </ui.Routed> 
```

And instead of adding `restricted: ...` to the tabs you can wrap the 
restricted components in `<Restricted>`.

norling:
That's a real neat solution! :) I'll try it!

danr (approved):
Great that it worked :)

jhagberg (approved):

PR #151, opened 2020-10-21
Feature/external dummy herds

### LINKED ISSUE #118: External Dummy Herds
Now animals that are sold outside a registred herd is still active in the founder herd. One way of solving this is by the comment below

andkaha:
Adding dummy herds `GX1` and `MX1` for the two genebanks.  These will hold the animals sold outside of the genebanks (by means of entries added to the `herd_tracking` table).

### DIFF
--- a/frontend/src/user_context.tsx
+++ b/frontend/src/user_context.tsx
@@ -99,7 +99,7 @@ export function WithUserContext(props: {children: React.ReactNode}) {
       }
     }
     // id is a herd
-    else if (/^([a-zA-Z][0-9]+)$/.test(id)) {
+    else if (/^([a-zA-Z]X?[0-9]+)$/.test(id)) {
       const genebank = genebanks.find(genebank => genebank.herds.some(h => h.herd == id))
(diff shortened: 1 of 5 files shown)


norling (changes requested):
If we're going to allow multiple letters in Herd prefixes you need to update the regexps in `frontend/src/user_context.tsx: canEdit()` so that they still trigger. This probably won't be an issue for the dummy herds, but if we allow these herds then the code should reflect that.

andkaha:
> If we're going to allow multiple letters in Herd prefixes you need to update the regexps in `frontend/src/user_context.tsx: canEdit()` so that they still trigger. This probably won't be an issue for the dummy herds, but if we allow these herds then the code should reflect that.

This is addressed in commit  7f164a4.

andkaha:
Dummy herds now added, but data for dummy herds is only available for Gotlandskanin as Mellerudskanin does not annotate animals sold to herds outside of the genebank.

jhagberg (approved):
LGTM good with name "Externa djur"

norling (approved):

PR #154, opened 2020-10-26
Feature/multi stage main

jhagberg:
Change so the API is part of the main docker file instead of separate container that is started in the compose

### DIFF
--- a/.docker/main
+++ b/.docker/main
@@ -1,13 +1,11 @@
-FROM python:3.8-buster as herdbook_api
+FROM python:3.8-buster
 
-RUN mkdir -p code
+RUN mkdir -p /code
 
 COPY app/ /code/
 WORKDIR /code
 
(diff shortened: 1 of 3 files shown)


.docker/main
```diff
@@ -1,3 +1,12 @@
+FROM python:3.8-buster as herdbook_api
+
+RUN mkdir -p code
+
+COPY app/ /code/
+WORKDIR /code
+
+RUN pip3 install -r /code/requirements-pydigree.txt && pip3 install -r /code/requirements.txt && chown -R www-data:www-data /code/
+
 FROM herdbook_api:latest
```
pontus: With `:latest` here, this will be taken as an image reference and docker will thus try pulling it if it's not available locally.

But I think we can in fact  just remove the `as herdbook_api` in line 1 and the `FROM herdbook_api:latest`  (entire line 10).

jhagberg: Yes that should work also. For me this separation was just confusing but if there is no need to have a api image stand alone it can just be one file!

pontus (approved):

PR #156, opened 2020-10-26
Added workflow to publish container packages

pontus:


jhagberg:
so What will this do build and publish images on push to Develop?

pontus:
Current behaviour is to build for all pushes. Tagged as the last part of the branch name if not in a special branch whereas for main, master and develop it will tag with latest instead.
 
The intent is that this should help with caching user side as well, but some more investigation is required there.

jhagberg (approved):
Works but needs some hack to get the cache locally. Need documentation

pontus:
Merged this now and will continue in another PR.

PR #157, opened 2020-10-27
Feature/herd editing

norling:
Mostly a big form, and a big form parsing function.

Adding new individuals will have to wait until Andreas is back so that we can agree on the database schema changes.

### DIFF
--- a/frontend/src/data_context.tsx
+++ b/frontend/src/data_context.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 import {get} from './communication'
 
-import {Colour, DataContext, Genebank, NameID} from "@app/data_context_global"
+import {Color, DataContext, Genebank, NameID} from "@app/data_context_global"
 
 /**
  * The data context holds genebank, herd, and individual data, as well as
@@ -21,7 +21,7 @@ export function useDataContext(): DataContext {
 export function WithDataContext(props: {children: React.ReactNode}) {
   const [genebanks, setGenebanks] = React.useState([] as Array<Genebank>)
   const [users, setUsers] = React.useState([] as Array<NameID>)
-  const [colors, setColors] = React.useState({} as {[genebank: string]:Colour[]})
+  const [colors, setColors] = React.useState({} as {[genebank: string]:Color[]})
 
   async function fetchAndSet(url: string, set: Function, field: string | undefined = undefined) {
     return await get(url).then(
--- a/frontend/src/data_context.tsx
+++ b/frontend/src/data_context.tsx
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 import {get} from './communication'
 
-import {DataContext, Genebank, NameID} from "./data_context_global"
+import {Colour, DataContext, Genebank, NameID} from "@app/data_context_global"
 
 /**
  * The data context holds genebank, herd, and individual data, as well as
@@ -21,6 +21,7 @@ export function useDataContext(): DataContext {
 export function WithDataContext(props: {children: React.ReactNode}) {
   const [genebanks, setGenebanks] = React.useState([] as Array<Genebank>)
   const [users, setUsers] = React.useState([] as Array<NameID>)
+  const [colors, setColors] = React.useState({} as {[genebank: string]:Colour[]})
 
   async function fetchAndSet(url: string, set: Function, field: string | undefined = undefined) {
     return await get(url).then(
@@ -88,7 +89,7 @@ export function WithDataContext(props: {children: React.ReactNode}) {
         }
         // note that individuals aren't loaded, and send a request to load them
         const genebankData = data['genebanks'].map((g: Genebank) => {
-          g.individuals = null
+          g.individuals = []
           return g
         })
         setGenebanks(genebankData)
@@ -115,6 +116,12 @@ export function WithDataContext(props: {children: React.ReactNode}) {
   async function getUsers() {
     return fetchAndSet('/api/manage/users', setUsers, 'users')
   }
+  /**
+   * Fetches all legal colors for all genebanks.
+   */
+  async function getColors() {
+    return fetchAndSet('/api/colors', setColors)
+  }
 
   /**
    * Loads data from the backend into the data context. Returns `true` if all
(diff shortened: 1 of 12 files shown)


frontend/src/data_context.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 import {get} from './communication'
 
-import {DataContext, Genebank, NameID} from "./data_context_global"
+import {Colour, DataContext, Genebank, NameID} from "@app/data_context_global"
```
danr: Uh oh, this file uses both `Colour` and `Colors`. I would go for one of the spellings, AE in particular.

frontend/src/message_context.tsx
```diff
@@ -40,7 +40,8 @@ export function WithMessageContext(props: {children: React.ReactNode}) {
   const [dialogLink, setDialogLink] = React.useState(undefined as string | undefined)
 
   function userMessage(message: string, severity: MessageLevel) {
-      setMessage(message);
+      // print json format if the message isn't a string.
+      setMessage(typeof message == "string" ? message : JSON.stringify(message));
```
danr: Use `JSON.stringify(message, undefined, 2)` for some indentation

danr (approved):
Looks good! I haven't actually tried running this, just read the code.

One thing I noticed in encam was that Autocompletes are quite slow to render so I had to use some memo to improve the performance. Maybe won't be a problem on this page.

:clinking_glasses:

danr:
Ok another comment: I was a bit surprised with that you store the string representation of them in the state rather than the actual Date object. I haven't actually worked with dates much so I don't know what's most convenient though.

norling:
yeah - the string representation is a bit dumb, but I need to have a set format, otherwise the backend parser gets confused... python doesn't seem to have a "guess date format" setting, and javascript dates default to strings like `Tue Oct 27 2020 12:54:46 GMT+0100 (centraleuropeisk normaltid)` which are annoying to parse...

frontend/src/data_context.tsx
norling: yeah... it's mostly a nod to Andreas always using "colour" :) And the british spelling never collide with any other definitions ^^
I should probably change it to `Color` again though...

frontend/src/data_context.tsx
```diff
@@ -115,6 +116,12 @@ export function WithDataContext(props: {children: React.ReactNode}) {
   async function getUsers() {
     return fetchAndSet('/api/manage/users', setUsers, 'users')
   }
+  /**
+   * Fetches all legal colors for all genebanks.
+   */
+  async function getColors() {
+    return fetchAndSet('/api/colors', setColors)
+  }
```
jhagberg, line 124: I tested now and in color i only get "no option" Should this be populated by allowed colors for Gotlands kanin?

frontend/src/individual_edit.tsx
```diff
@@ -0,0 +1,424 @@
...
+              <TextField label='Mätningsdatum' className={style.control}
+                value={weightDate}
+                type='date'
+                variant={inputVariant}
+                InputLabelProps={{
+                  shrink: true,
+                }}
+                onChange={(e: any) => {setWeightDate(e.target.value)}}
+                />
+              <TextField label='Mätningsdatum' className={style.control}
```
jhagberg:
```suggestion
              <TextField label='Vikt' className={style.control}
```
Both fields have the same label should this be Vikt instead?

frontend/src/individual_edit.tsx
```diff
@@ -0,0 +1,424 @@
...
+                      Delete
+                    </a>]
+                  </span>
+                </li>)}
+            </ul>
+            <div className={style.flexRow}>
+              <TextField label='Mätningsdatum' className={style.control}
+                value={weightDate}
+                type='date'
+                variant={inputVariant}
```
jhagberg, line 349:
```suggestion
                variant={inputVariant}
                InputProps={{
                 endAdornment: <InputAdornment position="end">Kg</InputAdornment>
                   }}

```
Could this be something to add so we know what Unit to write the weights in

frontend/src/individual_edit.tsx
```diff
@@ -0,0 +1,424 @@
...
+ * add a new individual if `herdId` is given.
+ */
+import React from 'react'
+import { makeStyles } from '@material-ui/core/styles';
+import { get, update } from '@app/communication';
+import { BodyFat, DateBodyfat, DateWeight, Individual, individualLabel,
+         LimitedIndividual, OptionType, ServerMessage,
+        } from '@app/data_context_global';
+import { useMessageContext } from '@app/message_context';
+import { Button, CircularProgress, TextField,
```
jhagberg:
```suggestion
import { Button, CircularProgress, TextField, InputAdornment,
```
Need to Import InputAdornment to use Kg in prefix in input?
I am not sure what is best here. Maybe @danr  Nows a better way to do this

jhagberg (changes requested):
I have added some suggestions.

And also I got new information from Föreninge
"Spontant, för att det inte ska bli totalt kaos, föreslår jag att alla fält ska vara låsta, utom fälten för köpare, där den som utfärdat intyget senare kan skriva in köpare och kontaktuppgifter till denna. " 

This is added in #161 

So Basically regular user cant edit any fields of a rabbit just add weights and hull information And add notes about the rabbit.

frontend/src/individual_edit.tsx
```diff
@@ -0,0 +1,424 @@
...
+  {individual
+    ? <div className={style.form}>
+        <div className={style.flexRowOrColumn}>
+          <div className={style.formPane}>
+            <div className={style.titleText}>
+              Redigera Individ
+            </div>
+            <div className={style.adminPane}>
+              <div className={style.paneTitle}>
+                Admin Fields
```
jhagberg:
```suggestion
                Kan bara ändras av Genbanksansvarig
```
Can we also have this to expand to the current Genebank Owners name?

frontend/src/data_context.tsx
norling: that's so weird, but the `/api/colors` endpoint disappeared. I re-added it, but I have no idea where it went...

frontend/src/individual_edit.tsx
norling: yes! of course! good catch!

frontend/src/individual_edit.tsx
jhagberg: The second part is low prio and added in #159

jhagberg (approved):
Think this is OK 
Adresses things in 
#161 
#160 

and low prio feature #159

PR #162, opened 2020-10-27
Feature/permission fields

### LINKED ISSUE #30: As a herd owner I want to be able to select what information other herds can see on my herd page
I would like to be able to select what information other herds can see about my herd.
Also, it should be possible to make a "public" page where I can list rabbits for sale to make it easier for new members to find rabbits to found there own herd with 

_(This could simply be solved by the herd owner themself not inputting any information.)_

norling:
Adds permission selectors to all herd contact person fields.

Closes #30

### DIFF
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -59,6 +59,9 @@ export interface Individual extends LimitedIndividual{
     alive: boolean
 }
 
+export type PrivacyLevel = 'private' | 'authenticated' | 'public' | null;
+
+
 export interface Herd {
     id: number
     genebank: number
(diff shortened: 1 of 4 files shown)


frontend/src/data_context_global.tsx
```diff
@@ -59,6 +59,9 @@ export interface Individual extends LimitedIndividual{
     alive: boolean
 }
 
+export type PrivacyLevel = 'private' | 'authenticated' | 'public' | null;
```
danr, line 62: I would change `null` to a string describing what privacy level it stands for. I don't know what `null` is here :)

frontend/src/field_with_permission.tsx
```diff
@@ -0,0 +1,61 @@
...
+});
+
+/**
+ * The FieldWithPermission function provides an input field, with the input tag
+ * `field`, having the input label `label`, with the value `value`. If
+ * permission is not `null`, then a permission field will be attached with the
+ * level `permission`. When a value is changed in either field the new value
+ * will be returned by the `setValue` callback.
+ */
+export function FieldWithPermission({field, label, value, permission, setValue}:
```
danr, line 34: I think here I would just call them `props`and then refer to them as `props.field`, `props.label` and so on since they are used at few locations in the body.

danr (approved):
Looks good! Feel free to ignore my comments if you disagree.

jhagberg (approved):
Tested seams to work good. Maybe need some polish UI vise but it works so 
Look at the Dans comments or merge it

frontend/src/data_context_global.tsx
norling: null is quite literally when the value is NULL in the database.

PR #167, opened 2020-10-28
Display F and MK on Individual view

jhagberg:
Added the values of F and MK on the individual view
Using Tooltip to explain the meaning of them

Would like to display the tooltip for both the F: and the actual value.


danr (approved):
Looks ok

PR #168, opened 2020-10-28
Feature/limit owner permissions

### LINKED ISSUE #161: Most fields should not be editable on a "registered" rabbit
If a rabbit is registered and put in the "active" population and got a certificate number. The fields should only be editable by the Genebank owner and not by herd owner. 

Herd owner should be able to add a note. Add weights and hull. And be able to sell the rabbit to External herd or another internal hed

norling:
Makes sure that owners can only edit notes in the individual editing form.

Closes #161

### DIFF
--- a/frontend/src/individual_edit.tsx
+++ b/frontend/src/individual_edit.tsx
@@ -248,7 +248,7 @@ export function IndividualEdit({id}: {id: string | undefined}) {
                 onChange={(event) => {updateField('number', event.currentTarget.value)}}
               />
               <TextField
-                disabled={!canManage}
+                disabled={!(isNew || canManage)}
                 label="Certifikat"
                 className={style.control}
                 variant={inputVariant}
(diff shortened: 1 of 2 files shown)


frontend/src/individual_edit.tsx
```diff
@@ -241,7 +248,7 @@ export function IndividualEdit({id}: {id: string | undefined}) {
                 onChange={(event) => {updateField('number', event.currentTarget.value)}}
               />
               <TextField
-                disabled={!user?.canEdit(individual.genebank)}
+                disabled={!canManage}
```
danr: No need to check `isNew` here?

danr (approved):
Looks good

frontend/src/individual_edit.tsx
jhagberg: Yes a herd owner might want to input a new rabbit In the system that use a old system aka a paper certificate and if so they would like to input the certificate number from the physical paper I'm the system.
If the the herd owner want to issue a digital certificate of the New rabbit the certificate number should come from our system. So possible three different number series of the certificate number.

jhagberg (changes requested):
Herd owner might want to input certificate numbers of they write a paper certificate for the rabbit and not issue a digital certificate from the our system.

frontend/src/individual_edit.tsx
norling: ok! :) I had understood things as the certificate always coming from the manager, but that's an easy fix!

jhagberg (approved):
Good

PR #172, opened 2020-10-30
Feature/test branch script

jhagberg:
Do not know if this will be of any help but it speeds up testing branches that have prebuilt images !


README.md
```diff
@@ -68,3 +68,9 @@ To be able to login in the website and play with it you will need to create an u
 ./register_user.sh 'user@domain.com', 'userpassword'
 ```
 
+If you just want to test a PR or a branche that is pushed to Github we now have a Github Action that will build the images
+So to just pull and test the prebuilt images use the script
```
norling: misspelled "branch". Also - I think it might be clearer written as:
```
All branches that are pushed to github has prebuilt images. To use the prebuilt images instead of building locally, use:
```
or something like that :)

dc-current-branch.yml
```diff
@@ -0,0 +1,43 @@
...
+        image: postgres:12
+        container_name: herdbook-db
+        env_file:
+            - ./.docker/database-variables.env
+        environment:
+            - POSTGRES_HOST_AUTH_METHOD=trust
+        volumes:
+            - ./postgres-data:/var/lib/postgresql/data
+    herdbook-frontend-devel:
+        image: ghcr.io/nbisweden/herdbook_frontend:${TAG}
```
norling: It might be neat to use the `${TAG:?error-message}`-syntax to give a helpful error message if TAG isn't set properly!

test-current-branch.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+GITHUB_REF="$(git rev-parse --abbrev-ref HEAD)" TAG=${GITHUB_REF##*/} docker-compose -f dc-current-branch.yml up
```
norling: newline isn't a more expensive character than space.

norling (changes requested):

README.md
jhagberg: Make a suggestion and commit! :)

norling (approved):

pontus (approved):

pontus:
Could possibly add the option of passing a branch to run for instead of the current, but this is very good as is.

jhagberg:
Yes we talked about that it could be good but we can just do an new PR! :)

PR #173, opened 2020-10-30
Feature/admin herd filter

### LINKED ISSUE #93: Funderingar 2
2020-07-29
Administrera besättningar
Det bör finnas en funktion där genbanksansvariga noterar ett val: Aktiva genbanker / Avslutade genbanker. För avslutade genbanker ska kontaktuppgifterna i samband med avslutet raderas (av genbanksansvarig). Bra kanske också med en funktion där man kan välja ”Visa bara aktiva genbanker”
De flesta genbankerna för gotlandskanin är faktiskt avslutade. I dagsläget är 477 avslutade och 246 aktiva

norling:
Another small thing - option to filter inactive herds from the manager herd list.

Closes #93

### DIFF
--- a/frontend/src/manage.tsx
+++ b/frontend/src/manage.tsx
@@ -9,10 +9,12 @@ import { Switch, Route, useLocation, useHistory } from "react-router-dom";
 
 import { useDataContext } from './data_context'
 import Select from 'react-select';
-import { Genebank, NameID, Herd, herdLabel, userLabel } from '@app/data_context_global';
-import { Button } from '@material-ui/core';
+import { Genebank, NameID, Herd, herdLabel, userLabel, OptionType
+        } from '@app/data_context_global';
+import { Button, Checkbox, FormControlLabel, TextField } from '@material-ui/core';
 import { HerdForm } from '@app/herdForm';
 import { UserForm } from '@app/userForm';
+import { Autocomplete } from '@material-ui/lab';
 
 const useStyles = makeStyles({
   main: {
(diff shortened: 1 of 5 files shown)


jhagberg (commented):
Make all AutoComplete controlled What does that commit do?

jhagberg (approved):

norling:
material-ui's AutoComplete widgets inform the react state handler that they are "controlled" or "uncontrolled" when they are first rendered - this is done by checking if the value isn't `undefined`, so this commit adds a fallback to make sure that the value can never be undefined.

frontend/src/manage.tsx
```diff
@@ -120,6 +125,20 @@ export function Manage() {
...
+  const filtered = (options: OptionType[]) => {
+    if (showInactive || topic == 'user') {
+      return options;
+    }
+    const genebank = genebanks.find(g => g.name == topic)
+    if (!genebank) {
+      return options;
+    }
+    return options.filter(option => {
+      const herd = genebank.herds.find(h => h.herd == option.value)
```
danr: O(n²), how many herds do we have? If they're too many this will become sluggish. But say <100 should be no problem.

frontend/src/manage.tsx
```diff
@@ -58,12 +60,15 @@ export function Manage() {
...
   const [selected, setSelected] = useState(undefined as any)
+  const [showInactive, setShowInactive] = useState(false)
 
   /**
    * Set the options of the main select box to the list of current users.
    */
   const setUserOptions = () => {
-    const userOptions: NameID[] = users.map((u: NameID) => {return {value: u.id, label: userLabel(u)}});
+    const userOptions: OptionType[] = users.map((u: NameID) => {
+      return {value: `u.id`, label: userLabel(u)}
```
danr:
```suggestion
      return {value: u.id, label: userLabel(u)}
```

danr (changes requested):

frontend/src/manage.tsx
norling: yeah - it needs to be `'' + u.id` (or similar cast to string), I noticed that too ^^

danr (approved):

danr:
Cool bananas! :banana: :clinking_glasses:

PR #175, opened 2020-11-05
Feature/big table improvements

### LINKED ISSUE #130: Order herd select fields by number
Select fields that display herds should be ordered by number and not alphabetically.

### LINKED ISSUE #132: Allow a subset of columns by default in FilterTable
Currently the FilterTable crashes if you add columns that weren't in there when the Table is created.

norling:
This still doesn't allow column sorting, but it handles the rest of the table a bit better, and doesn't need the material-table dependency anymore. I'm gonna keep looking at the column sorting, but this started to grow to large, so I figured that it was better to make a PR with just the new table!

Closes #130 
Closes #132

### DIFF
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -13,7 +13,7 @@ import { CircularProgress, Checkbox,  makeStyles, FormControlLabel, TextField
         } from '@material-ui/core';
 import { Autocomplete } from '@material-ui/lab';
 
-import { asLocale, Individual, OptionType } from '@app/data_context_global';
+import { asLocale, Individual, inputVariant, OptionType } from '@app/data_context_global';
 import { IndividualView } from '@app/individual_view';
 import { HerdView } from './herd_view'
 import { useMessageContext } from '@app/message_context';
@@ -180,7 +180,10 @@ export function FilterTable({individuals, title = '', filters = [],
         value={columns.filter((v: any) => !v.hidden).map((v: any) => {return {value: v.field, label: v.title}})}
         getOptionLabel={(option: OptionType) => option.label}
         getOptionSelected={(option: OptionType, value: OptionType) => option.value == value.value}
-        renderInput={(params) => <TextField {...params} label='Kolumner' margin="normal" />}
+        renderInput={(params) => <TextField {...params}
+          label='Kolumner'
+          variant={inputVariant}
+          margin="normal" />}
         onChange={(event: any, newValues: OptionType[] | null) => {
           newValues && updateColumns(newValues)
         }}
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -144,8 +144,8 @@ function individualSort(a: Individual, b: Individual,  column: Column,
         bVal = f(bVal)
       }
       else if (column.sortAs == 'date') {
-        aVal = new Date(aVal)
-        bVal = new Date(bVal)
+        aVal = aVal ? new Date(aVal) : 0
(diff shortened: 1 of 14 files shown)


jhagberg (commented):
Get a snowpack error on the dev server
```
[snowpack] installing dependencies...
herdbook-frontend-devel_1  | [snowpack] Package "@date-io/date-fns" not found. Have you installed it? 

```

jhagberg (changes requested):
Things I notice when testing Looks nicer now I think. Search is a bit better looks like it renders faster also.

problems or questions:

- Cant hide Dead animals on Herd page. 
- The death date is set to "11/5/2020" for all animals both in hed view and in genebank view.
- Opening a herd on new page gives nginx error 404

frontend/src/filter_table.tsx
```diff
@@ -72,74 +40,112 @@ const useStyles = makeStyles({
     color: 'blue',
     textDecoration: 'underline',
     cursor: 'pointer',
+  },
+  hidden: {
```
danr: What's happening here? Why not `display: none` or `visibility:hidden`?

frontend/src/filter_table.tsx
```diff
@@ -165,25 +174,150 @@ export function FilterTable({individuals, title = '', filters = [],
...
+   * subfield to use for sorting. `sortAs` can be `number`, `numbers` or `date`
+   * to force a column to be sorted as a particular type.
+   * - `date` fields will be cast as dates using `new Date(<value)`
+   * - `number` fields will be cast as numbers by removing all non-numerical
+   *   characters. ex. `G111-222` -> `111222`
+   * - `numbers` fields are cast as numbers, but using the first set of numbers
+   *   as the integer part, and the remaning as decimals.
+   *   ex. `G111-222-333` -> `111.222333`.
+   */
+  const sortedIndividuals = React.useMemo(() => {
```
danr, line 302: I would go for putting the sort function on the top level or in a separate module so it's easier to test.

frontend/src/filter_table.tsx
```diff
@@ -72,74 +40,112 @@ const useStyles = makeStyles({
...
+  search: {
+    float: 'right',
   }
 });
 
+interface Column {
+  field: keyof Individual;
+  label: string;
+  sortBy?: string;
+  sortAs?: string;
```
danr: A bit unsure what the difference between sortBy and sortAs are here

frontend/src/filter_table.tsx
```diff
@@ -155,7 +161,10 @@ export function FilterTable({individuals, title = '', filters = [],
     setColumns(newColumns)
   }
 
-  const filteredIndividuals = () => {
+  /**
+   * Memoized list holding all individuals that pass the current filter(s).
+   */
+  const filteredIndividuals: Individual[] = React.useMemo(() => {
```
danr, line 280: I would put this in a top-level function as well to make testing easier.

danr (approved):
Some comments on the shape of the code

frontend/src/filter_table.tsx
norling: ah - should have commented on that. `sortBy` is to sort the column by a sub-field when the value is an object, `sortAs` is to coerce the value to a new type before sorting!

frontend/src/filter_table.tsx
norling: ah! forgot to change the name ^^ That class is for the sorting arrow.

danr (approved):
Alright! :sunglasses:

jhagberg (approved):
LGTM

PR #177, opened 2020-11-06
Use / as last resort for non-existant URIs.

pontus:


jhagberg (changes requested):
This does not work just get me a blank page

pontus:
Odd, but this seems to work for me now at least.

jhagberg:
Running this is one step forward but still does not work. It now generates the root page on all subdirectory access. 
 For example "http://localhost:8080/genebank/Gotlandskanin" rewrites to /

jhagberg:
grejt this seams to work!

jhagberg (commented):
Looks like it works now

jhagberg (approved):

PR #179, opened 2020-11-06
Add basic validation to contact form

### LINKED ISSUE #163: Email field in Herd has no validation
Email field in Herd edit has no validation.

We need basic convenience validation in the frontend - mostly to help the user avoid typos like missing "@".
This only applies to the frontend, and is for the users convenience. 

Examples of solutions could be a regexp or using a validation package.
All valid email addresses don't need to pass the validation.

norling:
Add some basic validation to contact form.

This PR probably needs extra awesome review since I was really tired writing it.

Closes #163


frontend/src/field_with_permission.tsx
```diff
@@ -24,15 +24,46 @@ const useStyles = makeStyles({
...
+ * @param value
+ */
+function validateType(fieldType: LimitedInputType, value: string | null): boolean {
+  if (!value) {
+    return true
+  }
+  switch (fieldType) {
+    case 'email':
+      // Validation with RFC5322
+      return !!value.match(/(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])/);
```
danr: Why not use a dependency here? Perhaps https://www.npmjs.com/package/isemail

frontend/src/field_with_permission.tsx
```diff
@@ -24,15 +24,46 @@ const useStyles = makeStyles({
...
+    case 'url':
+      // validation from urlregex.com
+      return !!value.match(/((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/)
```
danr, line 31: Should be a library for this too

frontend/src/field_with_permission.tsx
```diff
@@ -24,15 +24,46 @@ const useStyles = makeStyles({
...
+      // very general match - basically allow anything that starts with
+      // `+<num>(0)` or `0`, followed by any combination of numbers, whitespace
+      // and dashes.
+      return !!value.match(/^(\+?[0-9]{1,3}\(0\)|0)[\s0-9\-]+$/)
+  }
+  return true
+}
+
+
+export type LimitedInputType = 'text' | 'tel' | 'url' | 'email';
```
danr, line 42: `tel` seems Swedish. Perhaps `tel` -> `phone_number` ?

frontend/src/field_with_permission.tsx
```diff
@@ -24,15 +24,46 @@ const useStyles = makeStyles({
...
+    case 'url':
+      // validation from urlregex.com
+      return !!value.match(/((([A-Za-z]{3,9}:(?:\/\/)?)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/)
+    case 'tel':
+      // very general match - basically allow anything that starts with
+      // `+<num>(0)` or `0`, followed by any combination of numbers, whitespace
+      // and dashes.
+      return !!value.match(/^(\+?[0-9]{1,3}\(0\)|0)[\s0-9\-]+$/)
+  }
+  return true
```
danr, line 38: If you support `case 'text'` the type-checker will notify you when you add more types and forget to add them to the switch to make it exhaustive.

danr (commented):
Some comments to consider

frontend/src/field_with_permission.tsx
norling: weirdly enough, `<input type='tel'>` is used for phone numbers... and I think it helps aria and mobile browsers, so I used the standard types :)

frontend/src/field_with_permission.tsx
norling: I guess we could - but does it provide any functionality that we need, that we don't get in this regexp?

frontend/src/field_with_permission.tsx
norling: same as above - would a dependency provide any benefit?

frontend/src/field_with_permission.tsx
danr: The benefit is that you won't need to test.

frontend/src/field_with_permission.tsx
danr: Jesus bananas

frontend/src/field_with_permission.tsx
jhagberg: What about that Formik library? or is that another use-case?

jhagberg: Formik and yup?

danr: yup yup has an email validator, they don't have any tests for it except that the empty string is not an email

norling: ah well...

norling: Formik email validation seems to be based on giving your own regexp though... and the one in their example will fail for tons of valid emails. Isemail seems to be equivalent to the regexp, but I can change it out since dependencies are so popular! :)

pontus (approved):

airenzp:
I see this old PR now, we should take a look at it I guess.

jonandernovella (approved):

PR #181, opened 2020-11-18
A docker environment for loading xlsx data in database using docker

jhagberg:
Added a small docker-compose file
A small container with everything needed to run the import scripts
Bash files with small changes to run everything using docker-compose
now re-run csv creation if file csv file is of zero bytes

### DIFF
--- a/scripts/load-docker-compose.sh
+++ b/scripts/load-docker-compose.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# This will use docker-compose to load the data
+# Using a stage built main container and the actula 
+# Postgres container
+
+# Usage example:
+#	./load-docker-compose.sh -g kanindata-gotland-v9.xlsx \
+#	          -G herd-registry-gotland.xlsx \
+#	          -m kanindata-mellerud-v4.xlsx
+
+usage () {
+	cat <<-USAGE_END
+	Usage:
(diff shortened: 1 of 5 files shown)


pontus:
I don't see the specific load scripts? (`load-mellerud.sh` and so on).

scripts/load-docker-compose.sh
```diff
@@ -0,0 +1,76 @@
...
+
+	    -h		Show this help text
+	USAGE_END
+}
+
+
+while getopts 'g:G:hm:' opt; do
+	case $opt in
+		g)	gfile=$OPTARG ;;
+		G)	Gfile=$OPTARG ;;
```
pontus, line 33: Naming? Could this be grabbit and gherd or grdata/ghdata or something?

pontus (approved):

pontus:
Approved now, but that is more to show that what's here is fine, I still think scripts are missing.

jhagberg:
The missing script is already in the repo I did not need to change them

scripts/load-docker-compose.sh
jhagberg: I just took @kusalananda script so need to update that instead

pontus:
Ah, must have confused branches or looked in an non-pulled repo.

PR #182, opened 2020-11-18
Update main to use multi-stage

jhagberg:
Maybe this will not speed up build so much
but it should be a bit faster to build

### DIFF
--- a/.docker/main
+++ b/.docker/main
@@ -1,22 +1,32 @@
-FROM python:3.8-buster
+FROM python:3.8-buster as main-base
 
-RUN mkdir -p /code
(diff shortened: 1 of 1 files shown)


norling (approved):
Couldn't test it much since my database is out of whack - but it builds faster and starts up fine! :)

jonandernovella (approved):

pontus:
This seems to disable the use of caching on my system (i.e. repeated `docker-compose build` will actually rebuild every time instead of using cached layers).

jhagberg:
aha thats is not good wonder why?

PR #183, opened 2020-11-19
Database updates to support "breeding" table

### LINKED ISSUE #147: Colors should be connected to genebanks in the database
Since not all colors are valid for all genebanks, this should be represented in the genebank so that illegal colors are not available when editing animals.

### LINKED ISSUE #166: Active as field in db
Should we have a field active in the individual table? 
Instead of calculating this field now, the genebank managers would like to update an animal status to set it active or inactive. We could also have a function to update this field if an animal is marked dead.

andkaha:
closes #166
closes #147


.gitignore
```diff
@@ -15,7 +15,11 @@ venv.bak/
 .vscode
 .tox/
 .docker/database-variables.env
-.docker/r-api-variables.env
```
jhagberg, line 18: Why this some old "skräp"?

jhagberg (approved):
Looks good otherwise. Have not tested anything

.gitignore
norling: dunno tbh. I think andreas removed it?
I doesn't seem to be needed anymore at least!

run_tests.sh
```diff
@@ -0,0 +1,19 @@
...
+
+if [ ! -d "$VENV" ]
+then
+  echo "Error: Could not find virtual environment '$VENV'."
+  echo "Create the environment with:\n\n    python -m venv $VENV.\n"
+  exit 1
+fi
+
+source "$VENV/bin/activate"
+
```
inghylt, line 19: add pip install -r app/requirements.txt ? :)

inghylt (approved):
Very nice! 

Two thoughts:
- test_api.py is quite large. Could be beneficial to separate some of the test classes into separate files
- Maybe add some info the root README.md about the tests?

PR #185, opened 2020-11-30
Use absolute paths in build.sh

danr:
I think this should be enough


pontus (approved):
Seems reasonable.

jhagberg (approved):
LGTM lets merge. Works when testing!

PR #186, opened 2020-12-02
Add script to dump/load container db

inghylt:


scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
+
+# some config variables
+OUTPUT_DIR="db_dumps"
+BASE="db-dump"
+TIMESTAMP=$(date +"%Y-%m-%d_%H%M")
+ENV_FILE=../.docker/database-variables.env
```
norling, line 18: It's mostly a matter of personal preference, but I like doing `cd $(dirname $0)` in scripts that need to do follow relative paths. This way it doesn't matter if you run the script from the same directory.

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
...
+
+# load database variables
+if [ ! -e "${ENV_FILE}" ]
+  then
+    echo "Couldn't find env file '$ENV_FILE'." >&2
+    exit 1
+  fi
+
+source ../.docker/database-variables.env
+
```
norling, line 58: if you want you could check that the variables aren't empty with `[ -z "$var" ]`, that way you can give more useful error messages when some environment variable is missing.

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
...
+      FLAGS="--clean $FLAGS"
+    fi
+  fi
+  echo "Restoring database from dump: ${DUMP_FILE}"
+  cat "$DUMP_FILE" | docker exec -i "${CONTAINER}" pg_restore $FLAGS
+
+else
+  if [[ "$1" == "dump" ]]
+  then
+    if [[ -z "$2" ]]
```
norling: you set OUTPUT_DIR at the top, so I'd default to that. If you want you can do `OUTPUT_DIR="${2:-db_dumps}"` to use `$2`, or default to `db_dumps` if `$2` is empty.

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
...
+    then
+      echo "Please specify which directory to put dump file into"
+      exit 1
+    fi
+    OUTPUT_DIR=$2
+    if [ ! -e "$OUTPUT_DIR" ]
```
norling: you can check that it exists and is a directory with `-d`!

norling (approved):
Some comments, but nothing major, so if you don't agree I'm fine with using this!

scripts/dump_load_db.sh
inghylt: Nice catch! I forgot to remove that config variable when adding outpur dir as an argument. But maybe is is better to have a default output dir and give users the possibility to specify an output dir, i.e. not have it as a required argument.

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
...
+    then
+      echo "Creating dump output directory"
+      mkdir $OUTPUT_DIR
+    fi
```
pontus: As a personal preference, I like `mkdir -p SOMETHING` instead of things like these (the error message is also informative enough).

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,89 @@
+#!/bin/bash
```
pontus: Possibly run with `-e` to get automatic aborts if something does not work as expected?

pontus (approved):

scripts/dump_load_db.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash -e
+
+cd $(dirname $0)
+
+# some config variables
+OUTPUT_DIR="$(realpath ${2:-../db_dumps})"
```
norling: This doesn't seem to work in OS X where `realpath` isn't standard. 

Maybe something like:
```suggestion
function path {
  if command -v realpath >/dev/null
  then
    realpath $1
  else
    echo "$1"
  fi
}

OUTPUT_DIR="$(path ${2:-../db_dumps})"
```

PR #192, opened 2020-12-11
Handling of external authentication

pontus:
Add ability to use external authenticators (OAuth2 to twitter and Google included) and link accounts.


danr:
I think there must be some better way than a timeout at `waitForExternalLogin`. Waiting seems like some kind of way of polling. 

In `waitForExternalLogin` you extract the returned user by trimming the text in the body. There surely must be some way to return a json object instead.

I don't understand how the rest of the page figures out if it is logged in or not.

I don't understand this comment:  `// We are not doing account linking`.

pontus:
> I think there must be some better way than a timeout at `waitForExternalLogin`. Waiting seems like some kind of way of polling.

It is polling, in a sense. I'd be happy for alternative suggestions. The one I can think of involves doing it in the same window add doing a redirect in the end.
 
> In `waitForExternalLogin` you extract the returned user by trimming the text in the body. There surely must be some way to return a json object instead.

Sure. Responses from the python are subject to change still (so far mimics what was there before).

> I don't understand how the rest of the page figures out if it is logged in or not.

As ar as I understand, this is handled by the bits looking at `user` (in `useEffect` and close to that).

> I don't understand this comment: `// We are not doing account linking`.

To make things less confusing on the server part (and what needs to be done at the external service), the server shares some things between linking and login. For this, we do an extra call to signal that we're not doing linking. (It's possible that can be improved somehow on the server side).

danr:
> As ar as I understand, this is handled by the bits looking at user (in useEffect and close to that).

I mean how (and where) is this user populated from information from the external login services?

pontus:
(I

> > As ar as I understand, this is handled by the bits looking at user (in useEffect and close to that).
> 
> I mean how (and where) is this user populated from information from the external login services?

Ah, that can probably be improved. For now it does a full reload and lets the rest of the app figure out it's logged in (which it seems to do by asking the server - `on_mount` in `user_context.tsx`).

[I'm somewhat torn about where the code is best placed.]

danr:
I think full reload is totally OK for external login. Can this be used to avoid doing the `window.open` and the polling?

pontus:
> I think full reload is totally OK for external login. Can this be used to avoid doing the `window.open` and the polling?

Nah, those are separate. It would rather be adding an optional parameter to the route(s) in react and having the server doing redirect to that address at the end (that would be kind of bad because it means we introduce requirements for the client to behave in certain ways).

jhagberg:
the organization now have a Google workspace would it possbile to only allow users using  a google acount from that workspace?

pontus:
Yes, and indeed without it ever being visible to the herdbook - when setting up the requisites for oauth2 with an organization, one can choose to have the application internal only, meaning it will return a failure if authentication is done with an account from another domain.

app/utils/database.py
```diff
@@ -1220,6 +1220,32 @@ def migrate_3_to_4():
         ).execute()
 
 
+def migrate_4_to_5():
+    """
+    Migrate between schema version 4 and 5.
```
jonandernovella: This will likely be a conflict.

app/herdbook.py
```diff
@@ -293,6 +342,144 @@ def login_handler():
     return get_user()
 
 
+@APP.route("/api/login/<string:service>", methods=["GET", "POST"])
+def external_login_handler(service):
+    """
+    Do an external authentication. The special sercvice
```
jonandernovella:
```suggestion
    Do an external authentication. The special service
```

jonandernovella (changes requested):

jonandernovella:
Would be nice to keep the user's full name in our database

app/utils/external_auth.py
```diff
@@ -0,0 +1,269 @@
...
+    # Only care about verified addresses
+    if not idtoken.get("email_verified", False):
+        return None
+
+    if "email" in idtoken:
+        return {
+            "email": idtoken["email"],
+        }
+
+    return None
```
jonandernovella, line 222: I think that here we could send the full name

app/herdbook.py
pontus: Fixed.

app/utils/database.py
pontus: Correct, it was, should be fixed now.

app/utils/data_access.py
```diff
@@ -126,6 +168,180 @@ def authenticate_user(name, password):
...
+
+    try:
+        with DATABASE.atomic():
+            authenticator = (
+                Authenticators.select()
+                .where(
+                    (Authenticators.auth == method)
+                    & (Authenticators.auth_data == ident)
+                )
+                .get()
```
jonandernovella, line 271: This could be a separate function. But I think it's fine.

jonandernovella (approved):
LGTM!

PR #193, opened 2020-12-11
Feature/breeding api

norling:
Creates the necessary API functions needed to get and create breeding events in the database.

Jonas - could you add docstrings to the remaining R-API endpoints?

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -420,10 +420,10 @@ def mean_kinship(g_id):
 
 
 def get_mean_kinship(g_id):
-     """
+    """
     Fetch the mean kinship matrix from R-api of the genebank given  by `g_id`.
     """
-    response = requests.get('http://{}:{}/meankinship/{}' \
+    response = requests.get('http://{}:{}/meankinship/{}'
                             .format(settings.rapi.host,
                                     settings.rapi.port,
                                     g_id),
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -185,6 +185,24 @@ def register_breeding():
     return jsonify(status)
 
 
+@APP.route("/api/birth", methods=["POST"])
+@login_required
+def register_birth():
+    """
+    Used to update breeding events in the database with birth information.
+    Returns a message formatted like:
+        JSON: {
+            status: 'success' | 'error',
+            message?: string
+        }
+    """
+    form = request.json
+    status = {"status": "error", "message": "Unknown request"}
+    if request.method == "POST":
+        status = da.register_birth(form, session.get("user_id", None))
+    return jsonify(status)
+
+
 @APP.route("/api/colors")
 @login_required
 def colors():
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -40,7 +40,7 @@ APP.config.update(
     CACHE_DEFAULT_TIMEOUT=300
 )
 
-#pylint: disable=no-member
+# pylint: disable=no-member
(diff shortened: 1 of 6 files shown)


app/herdbook.py
```diff
@@ -254,8 +308,8 @@ def individual(i_number):
         try:
             ind["inbreeding"] = "%.2f" % (get_ind_inbreeding(i_number, ind['genebank_id']) * 100)
             ind["MK"] = "%.2f" % (get_ind_mean_kinship(i_number, ind['genebank_id']) * 100)
-        except requests.exceptions.ConnectionError as e:
-            logging.error(f'{e}')
+        except requests.exceptions.ConnectionError as error:
+            logging.error('%s', error)
```
pontus: `logging` or `APP.logger`?

app/herdbook.py
```diff
@@ -293,74 +347,73 @@ def get_ind_inbreeding(i_number, g_id):
...
     if response.status_code == 200:
         return csvparser.parse_csv(response.content)
 
-    else:
-        APP.logger.error("Could not fetch mean kinship data.")
-        APP.logger.error("Error {}".format(response))
-        return {}
+    APP.logger.error("Could not fetch mean kinship data.")
+    APP.logger.error("Error {}".format(response))
+    return {}
```
pontus, line 437: Flag error clearer?

app/utils/database.py
```diff
@@ -366,6 +366,17 @@ class Breeding(BaseModel):
...
+        """
+        Returns the objects key/value pair as a dictionary, using parent numbers
+        instead of database id's.
+        """
+        data = super().as_dict()
+        #pylint: disable=no-member
+        data['mother'] = self.mother.number
+        data['father'] = self.father.number
+        return data
+
```
pontus: Can it be clarified why this is needed?

app/tests/test_endpoints.py
```diff
@@ -112,6 +115,102 @@ def test_get_users(self):
...
                 )
 
+    def test_breeding_list(self):
+        """
+        Checks that `herdbook.breeding_list` returns the intended information.
+        """
+
+        # not logged in
+        self.assertEqual(self.app.get("/api/breeding").get_json(), None)
+
```
inghylt, line 125: maybe test status code here as well?

inghylt (approved):

app/herdbook.py
norling: yes! good catch!

app/herdbook.py
norling: How do you mean? Also - I didn't write this function, I just cleaned it up, so I don't think I'm the right person to clarify things ^^

app/utils/database.py
norling: Sure! it's to get the number instead of database id from the foreign keys, but I'll add that to the docstring.

app/tests/test_endpoints.py
norling: We should go through all endpoints and make sure that they return reasonable html codes, but I think that's outside the scope of this PR, sadly :/

app/herdbook.py
pontus: It might be for the error issue rather, but it feels like this should flag an error clearer to the caller (but then again, maybe an empty message is that error).

app/utils/database.py
pontus: Sorry, but doesn't the as_dict from the super class do this already? (I haven't checked, but would expect so.)

PR #197, opened 2020-12-14
Update doc strings for the R-api

jhagberg:


pontus (approved):

pontus:
What standard are we using for R docstrings?

jhagberg:
no real standard :) it can be improved!

jhagberg:
Yes there is no real doc-strings in R but we could use a package and do something like this 
https://cran.r-project.org/web/packages/docstring/vignettes/docstring_intro.html

pontus (approved):

norling:
I was thinking that you should add docstrings for the R-API functions in `herdbook.py`. But the R scripts but this is also good! :)

PR #203, opened 2020-12-16
Css/basic layout

norling:
probably a bunch of fiddling left, but is this a good start or should I work on some other idea?


jhagberg (dismissed):
Esbuild build.sh scripts needs update to handle png jpg.
Should the images be put in a subfolder?

jhagberg:
@pontus I can get the images to load does this has something to do with Nginx?

e0:
@jhagberg I tried running this branch locally and the frontend container crashes with the error `snowpack: not found`. I suspect that's why the images are not loading either.

pontus:
Yep, this is nginx (over-)compensating for react. It also seems images are not copied into the main image as needed for them to be served, so I/someone still need to troubleshoot that.

pontus:
Pre-built images work for me now.

airenzp (approved):
Great improvement to the interface!!!
I added some minor changes

PR #204, opened 2020-12-17
Feature/breeding frontend

norling:
In an effort to keep PRs small - here's the breeding form on the herd page, but without save/update functionality (which will be in the next PR).

The FilterTable and SortedTable functions currently share a lot of code, and the plan is to merge these, but that was also postponed to a different PR.

### DIFF
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -101,7 +101,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
   return <>
     <form className={style.form}>
       <MuiPickersUtilsProvider utils={DateFnsUtils}>
-        <Typography variant='h6'>{data == 'new' && 'Nytt '}Avelstillfälle</Typography>
+        <Typography variant='h6'>{data == 'new' && 'Nytt '}Parningstillfälle</Typography>
         <div className={style.formBox}>
           <KeyboardDatePicker
             autoOk
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -143,7 +143,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
             }}
           />
           <TextField
-            label="Anteckningar om avelstillfället"
+            label="Anteckningar om parningstillfället"
             variant={inputVariant}
             className={style.wideControl}
             multiline
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -49,15 +49,17 @@ const emptyBreeding: Breeding = {
 export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: string | undefined}) {
   const style = useStyles();
   const { genebanks } = useDataContext()
-  const [form, setForm] = React.useState({...emptyBreeding} as Breeding)
-  const hasBirth = React.useMemo(() => {return !!form.birth_date}, [form])
+  const [formState, setFormState] = React.useState(emptyBreeding as Breeding)
+  React.useEffect(() => setFormState(!data || data == 'new' ? emptyBreeding : data),
+                 [data])
+  const hasBirth = React.useMemo(() => {return !!formState.birth_date}, [formState])
 
   /**
    * Sets a single key `label` in the `herd` form to `value` (if herd isn't
    * undefined).
    */
   const setFormField = <B extends keyof Breeding>(label: B, value: Breeding[B]) => {
-    form && setForm({...form, [label]: value})
+    formState && setFormState({...formState, [label]: value})
   }
   const herd: LimitedHerd | undefined = React.useMemo(() => {
     const herd = genebanks.map(g => g.herds.find(h => h.herd == herdId))
@@ -89,17 +91,8 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
     })
   }, [genebank])
 
-  React.useEffect(() => {
-
-    if (!data || data == 'new') {
-      setForm({...emptyBreeding})
-    } else {
-      setForm(data)
-    }
-  }, [data])
-
   return <>
-    <form className={style.form}>
+    <form className={style.formState}>
       <MuiPickersUtilsProvider utils={DateFnsUtils}>
         <Typography variant='h6'>{data == 'new' && 'Nytt '}Parningstillfälle</Typography>
         <div className={style.formBox}>
@@ -110,7 +103,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
             label="Parningsdatum"
             format={dateFormat}
             className={style.wideControl}
-            value={form ? form.breed_date ?? '' : ''}
+            value={formState ? formState.breed_date ?? '' : ''}
             InputLabelProps={{
               shrink: true,
             }}
@@ -118,7 +111,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
           />
           <Autocomplete
             options={allFemales ?? []}
-            value={allFemales.find(option => option.value == form.mother) ?? null}
+            value={allFemales.find(option => option.value == formState.mother) ?? null}
             getOptionLabel={(option: OptionType) => option.label}
             renderInput={(params) => <TextField {...params}
               label="Mor"
@@ -131,7 +124,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
           />
           <Autocomplete
             options={allMales ?? []}
-            value={allMales.find(option => option.value == form.father) ?? null}
+            value={allMales.find(option => option.value == formState.father) ?? null}
             getOptionLabel={(option: OptionType) => option.label}
             renderInput={(params) => <TextField {...params}
               label="Far"
@@ -148,7 +141,7 @@ export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: st
             className={style.wideControl}
             multiline
             rows={2}
-            value={form.breed_notes ?? ''}
+            value={formState.breed_notes ?? ''}
             onChange={(e: any) => {setFormField('breed_notes', e.target.value)}}
           />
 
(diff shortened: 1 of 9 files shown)


frontend/src/sorted_table.tsx
```diff
@@ -0,0 +1,311 @@
...
+ *   as the integer part, and the remaning as decimals.
+ *   ex. `G111-222-333` -> `111.222333`.
+ *
+ * @param a First object
+ * @param b Second object
+ * @param column Column definition of the table to be sorted
+ * @param orderBy sub-field to order by for object values
+ * @param order `asc` or `desc` for ascending or descending ordering
+ */
+function columnSort(a: any, b: any,  column: Column, orderBy: string, order: Order ) {
```
danr, line 85: Can we share code from `filter_table.tsx:individualSort`? It seems to be roughly the same

frontend/src/breeding_list.tsx
```diff
@@ -0,0 +1,116 @@
...
+      return genebank.individuals.filter(i => parentNumbers.includes(i.number))
+    })
+  }, [genebanks, breedingEvents])
+
+  const columns: Column[] = [
+    {field: 'breed_date', label: 'Parningsdatum', sortAs: 'date', hidden: false},
+    {field: 'breed_notes', label: 'Parningsanteckningar', hidden: true},
+    {field: 'mother', label: 'Moder', sortAs: 'numbers', hidden: false},
+    {field: 'father', label: 'Fader', sortAs: 'numbers', hidden: false},
+    {field: 'birth_date', sortAs: 'date', label: 'Födslodatum', hidden: false},
```
danr:
```suggestion
    {field: 'birth_date', sortAs: 'date', label: 'Födelsedatum', hidden: false},
```

This is used elsewhere in the page

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+      setForm({...emptyBreeding})
+    } else {
+      setForm(data)
+    }
+  }, [data])
+
+  return <>
+    <form className={style.form}>
+      <MuiPickersUtilsProvider utils={DateFnsUtils}>
+        <Typography variant='h6'>{data == 'new' && 'Nytt '}Avelstillfälle</Typography>
```
danr:
```suggestion
        <Typography variant='h6'>{data == 'new' && 'Nytt '}Parningstillfälle</Typography>
```
Är det ett parningstillfälle eller ett avelstillfälle?

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+    if (!data || data == 'new') {
+      setForm({...emptyBreeding})
+    } else {
+      setForm(data)
+    }
+  }, [data])
+
+  return <>
+    <form className={style.form}>
```
danr: How do we submit a new bunny session?

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+            format={dateFormat}
+            className={style.wideControl}
+            value={form ? form.breed_date ?? '' : ''}
+            InputLabelProps={{
+              shrink: true,
+            }}
+            onChange={(date, value) => {value && setFormField('breed_date', value)}}
+          />
+          <Autocomplete
+            options={allFemales ?? []}
```
danr, line 113: The autocomplete gets too slow with this many options... Perhaps need to fall back to a browser-builtin dropdown?

frontend/src/sorted_table.tsx
norling: as noted in the description - yeah, this code is very similar to FilterTable - The idea is to make FilterTable extend SortedTable in another PR! :)

frontend/src/breeding_form.tsx
norling: We use the button in the next PR! :) I'm sacrificing feature completeness for smaller PRs now... is that good or bad? it's sort of a problem that you can't really follow the entire process... but it IS smaller!

frontend/src/breeding_form.tsx
norling: yeah - I could just use a regular select thing... or even just a text field and validate that it's a known animal of the correct sex?

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+  return <>
+    <form className={style.form}>
+      <MuiPickersUtilsProvider utils={DateFnsUtils}>
+        <Typography variant='h6'>{data == 'new' && 'Nytt '}Parningstillfälle</Typography>
+        <div className={style.formBox}>
+          <KeyboardDatePicker
+            autoOk
+            variant="inline"
+            inputVariant={inputVariant}
+            label="Avelsdatum"
```
inghylt:
```suggestion
            label="Parningsdatum"
```

Parningsdatum?

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+              label="Far"
+              className={style.wideControl}
+              variant={inputVariant}
+              margin="normal" />}
+            onChange={(event: any, newValue: OptionType | null) => {
+              newValue && setFormField('father', newValue.value)
+            }}
+          />
+          <TextField
+            label="Anteckningar om avelstillfället"
```
inghylt:
```suggestion
            label="Anteckningar om parningstillfället"
```

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+        </div>
+        { data != 'new'
+          ? <>
+              <Typography variant='h6'>Födsel</Typography>
+              <div className={style.formBox}>
+                <KeyboardDatePicker
+                  autoOk
+                  variant="inline"
+                  inputVariant={inputVariant}
+                  label="Födslodatum"
```
inghylt:
```suggestion
                  label="Födelsedatum"
```

frontend/src/breeding_form.tsx
inghylt: I think it is nice with small PR:s! If you state that further functionality will be added, as you did with this PR, I think it makes sense to prioritize the PR size. Obv. personal view though :')

inghylt (approved):

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/**
+ * The BreedingForm function. This function allows users to create and update
+ * breeding events in the database.
+ */
+export function BreedingForm({data, herdId}: {data: Breeding | 'new', herdId: string | undefined}) {
+  const style = useStyles();
+  const { genebanks } = useDataContext()
+  const [form, setForm] = React.useState({...emptyBreeding} as Breeding)
```
danr: There is no point in copying the `emptyBreeding` object to the state. `emptyBreeding` should stay constant, it's a bug if this object is ever modified. (If you expect this happens you can debug it by using Object.freeze and Object.seal on emptyBreeding to make an error be thrown in this happens)

On the other hand, values passed to `setForm` shoud be copies, not mutated objects referenced elsewhere. Because if this would ever happen it could mean that you could make emptyBreeding get mutated.

Futher, the variable name `form` is a bit unfortunate because it's not the actual form, it's the state in the from. Up to you of course but I would go for `form_state` or simply `state` (because it's the main state of this component).

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+  const hasBirth = React.useMemo(() => {return !!form.birth_date}, [form])
```
danr:
```suggestion
  const hasBirth = !!form.birth_date
```
This is quick enough to not memo

frontend/src/breeding_form.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }, [genebank])
+
+  React.useEffect(() => {
+
+    if (!data || data == 'new') {
+      setForm({...emptyBreeding})
+    } else {
+      setForm(data)
+    }
+  }, [data])
```
danr: I use this technique to sync props to the state too and I think it works well. However, I would put it just under the `React.useState` defining `form` so that this syncing is apparent.

It can be a `useLayoutEffect` if the body of the callback is quick enough to avoid a redraw. This seems to be the case here.

frontend/src/breeding_form.tsx
danr: I see! That's cool :fist_oncoming:

frontend/src/sorted_table.tsx
danr: Alright!

frontend/src/sorted_table.tsx
```diff
@@ -0,0 +1,311 @@
...
+                                   ? () => {
+                                     if (selected == row.tableId) {
+                                       props.onClick(null)
+                                       setSelected(null)
+                                     } else {
+                                       props.onClick(row)
+                                       setSelected(row.tableId)
+                                     }
+                                   }
+                                   : () => {}}
```
danr, line 264:
```suggestion
                          onClick={props.onClick && () => {
                                     if (selected == row.tableId) {
                                       props.onClick(null)
                                       setSelected(null)
                                     } else {
                                       props.onClick(row)
                                       setSelected(row.tableId)
                                     }
                                   }}
```
I just use `&&` here, it's ok to pass undefined to event handlers

frontend/src/breeding_list.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  React.useEffect(() => {
+    if (id) {
+      get(`/api/breeding/${id}`).then(
+        (data: {breedings: Breeding[]}) => {
+          data && setBreedingEvents(data.breedings)
+        },
+        error => {
+          console.error(error)
+          userMessage(error, 'error')
```
danr, line 95: Perhaps the `get` from communication should call `userMessage` on errors so that you get error handling for free on all requests.

danr (approved):
Being nitpicky in the comments but overall it looks all good! :clinking_glasses:

frontend/src/breeding_form.tsx
danr: Sounds like a good idea if it's convenient enough to type in!

frontend/src/breeding_list.tsx
norling: Can I do that? I thought I couldn't import the message context since communication isn't part of the DOM hierarchy?

PR #207, opened 2020-12-18
Improve esbuild import.meta warning message

danr:
Seems like there is no way to remove the warning message but it can be improved to this:
```
src/main.tsx:46:0: warning: "import.meta" is not available in the configured target environment and will be empty
import.meta?.hot?.accept() // Dan: OK to ignore esbuild warning on this line
~~~~~~~~~~~
1 warning
```


norling (approved):
seems fair!

PR #213, opened 2020-12-27
Even more import fixes

pontus:
Hopefully, the last fixes required.


norling (approved):
didn't test but this seems very reasonable!

PR #214, opened 2021-01-08
Add support database schema migrations

pontus:
Templates and infrastructure to support database schema migrations.


norling (approved):
Looks good, and the tests pass just fine! :)

PR #219, opened 2021-01-15
Use chown with colon

danr:


pontus (approved):
Colon is the documented way to do this.

PR #220, opened 2021-01-19
Added herd import for Mellerud

pontus:


andkaha:
I need to test the actual loading of the Mellerud herd registry data. I haven't had time to do so yet... :-/

scripts/load-docker-compose.sh
```diff
@@ -13,67 +13,9 @@
...
-	echo 'Will not load extra Mellerud herd info'
-	load_mellerud_herds=false
-else
-    herd="-M /scripts/$Mfile"
-fi
-
-if docker-compose -f ../dc-db-load.yml ps | grep -q 'main-load-db.*\sUp\s.*'; then
-    # Alll good here
-    :
+if docker-compose -f ../dc-db-load.yml ps | grep -q 'main-load-db.*\sUp\s.*'
```
jhagberg: This should also check if the herdbook-db container is running or not

scripts/load-docker-compose.sh
```diff
@@ -82,7 +24,21 @@ else
     sleep 10
 fi
 
+# Add path "/scripts/" in front of each non-option argument.
+for opt do
+	case $opt in
```
pontus, line 28:
```suggestion
	case "$opt" in
```

scripts/load-docker-compose.sh
```diff
@@ -82,7 +24,21 @@ else
     sleep 10
 fi
 
+# Add path "/scripts/" in front of each non-option argument.
+for opt do
```
pontus, line 27: This is obviously ok, but my preference is for explicitness.

scripts/load-docker-compose.sh
andkaha: Will sort that out.

scripts/load-docker-compose.sh
andkaha: Quotes actually never needed here.

scripts/load-docker-compose.sh
andkaha: I wanted to avoid having two _identical_ command line parsing loops.  It's either that I add `/scripts/` in front of the pathnames in the wrapper script, or that I add them later in the actual loading script (`load-in-docker.sh`), in the `while` loop that parses the options.

andkaha:
> I need to test the actual loading of the Mellerud herd registry data. I haven't had time to do so yet... :-/

Now tested, and it seems to work well.

Need to fix that thing about making sure that the `herdbook-db` container is up and running.  I'll do that in this PR too since @jhagberg commented on it here.

scripts/load-docker-compose.sh
andkaha: (because `case` is a keyword, not a command, which is why you don't have to quote `$name` in `[[ -f $name ]]` either)

scripts/load-docker-compose.sh
andkaha: > This should also check if the herdbook-db container is running or not

Now fixed with the latest commit (5d9abef5adbf79fb2289efc49018ac853c29f1a0).

andkaha:
I'm ok with this PR a it is now. @pontus or @jhagberg to merge when they wish to do so.

pontus:
> I'm ok with this PR a it is now. @pontus or @jhagberg to merge when they wish to do so.

You are stil requested for review, can you approve if you are happy?

andkaha (approved):
Ok

andkaha:
> > I'm ok with this PR a it is now. @pontus or @jhagberg to merge when they wish to do so.
> 
> You are stil requested for review, can you approve if you are happy?

Sorry, done.

PR #227, opened 2021-03-04
Feature/test breed functionality

inghylt:
First draft of inbreed functionality. PR created for feedback, i.e. not ready to be merged with develop branch yet


frontend/src/inbreeding_form.tsx
```diff
@@ -0,0 +1,123 @@
...
+  bottomButton: {
+    float: 'left',
+    marginTop: '20px',
+  }
+});
+
+/* TODO, decide if type Indivial, useMemo and unpacking makes sense*/
+
+/* Filter genebank for active individuals of given sex*/
+function activeIndividuals(genebank: Genebank, sex: string){
```
norling: I prefer to have a docstring and a return type on all functions, but it's not important for the functionality.

frontend/src/inbreeding_form.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const activeFemales : Individual[] = activeIndividuals(genebank, 'female')
+
+  const activeMales : Individual[] = activeIndividuals(genebank, 'male')
+
+  const filterOptions = createFilterOptions<Individual>({
+  limit: 30,
+  });
+  
+  /* TODO, develop function to calculate COI and if there are sufficient generations*/
+  let COI = 3
```
norling: It might be good to not that COI = Coefficient of Inbreeding, as some developers might not be used to the biology stuff :)

frontend/src/inbreeding_form.tsx
```diff
@@ -0,0 +1,123 @@
...
+                  renderInput={(params) => <TextField {...params}
+                    label="Välj mor"
+                    className={style.wideControl}
+                    variant={inputVariant} />}
+          />
+          <Autocomplete
+                  options={activeMales}
+                  getOptionLabel={(option: Individual) => individualLabel(option)}
+                  value={male}
+                  // FEEDBACK, not sure if onChange "typing" solution optimal
```
norling: not sure what you mean here, but this seems fine!

frontend/src/inbreeding_recommendation.tsx
```diff
@@ -0,0 +1,50 @@
...
+ */
+import React from 'react'
+import { Individual, Genebank } from './data_context_global'
+import { useDataContext } from '@app/data_context'
+import { parentPedigree } from '@app/pedigree'
+import { PedigreeNetwork } from '@app/pedigree_plot'
+import { IndividualView } from '@app/individual_view'
+import { useMessageContext } from '@app/message_context';
+
+export function InbreedingRecommendation({female, male, COI, sufficientGenerations = true} : {female: Individual, male: Individual, COI: number, sufficientGenerations: boolean}) {
```
norling: Again - I prefer docstrings on everything, but Dan didn't like them, so you can do as you like! :)

frontend/src/inbreeding_recommendation.tsx
```diff
@@ -0,0 +1,50 @@
...
+  
+  let parents: Individual[] = [female, male]
+  
+  const pedigree = React.useMemo(() => parentPedigree(genebanks, parents, 4), [genebanks, parents])
+  const beneficialCOI = 5
+  const badCOI = 10
+  let breedCouple = `${female.name} och ${male.name}`
+  let recommendation
+  if (!sufficientGenerations) {
+    recommendation = <p> Too few generations available to make accurate assessment of inbreed coefficient </p>
```
norling: Mycket engelska/svengelska i alla rekommendationer, så kika om du kan skriva om dem.

frontend/src/navigation.tsx
```diff
@@ -62,6 +63,13 @@ export function Navigation() {
       visible: is_logged_in,
       icon: <AccountBalanceIcon />
     },
+    {
+      label: "Inavel",
```
jhagberg: This should be named something else. "Provparning"

frontend/src/navigation.tsx
```diff
@@ -62,6 +63,13 @@ export function Navigation() {
...
+      path: "/inavel",
```
jhagberg: same her /provparning is better I thin

frontend/src/pedigree.tsx
```diff
@@ -32,6 +32,22 @@ export function unique(xs: any[], key: string | undefined = undefined): any[] {
   })
 }
 
+// FEEDBACK, adapt method unique instead to return the commonAncestors as well as the unique nodes?
+export function commonAncestors(xs: any[], key: string | undefined = undefined): Set<any> {
```
norling: I don't quite understand what you want feedback on here?
also - `xs`, and `key` are very generic parameter name, so you should probably have a docstring to clarify what they are.

jhagberg (changes requested):
We need to be able to select what Genebank to test mate with

It would be good if we could do some more fancy interfaces for selecting breeders. 

Also, we can do something similar as "compare product" that you can select rabbits from differents herds and save them to a test mating interface somehow.

frontend/src/pedigree.tsx
```diff
@@ -243,3 +259,121 @@ export function herdPedigree(genebanks: Genebank[], herdId: string | undefined,
...
 }
+
+/**
+ * Return the edges that connect ancestor id to seeked descendant id
+ * @param edges the edges that make up the connections between the nodes in the pedigree
+ * @param ancestorId the ancestor id
+ * @param seekedDescendantId the child id
+ */
+export function getConnectingEdges(edges: Edge[], ancestorId:string, seekedDescendantId:string): Set<string> {
+  const getChildren = (edges: Edge[], parentId: string, seekeDescendantId: string) => {
```
norling: `seekeDescendantId` is probably misspelled.

frontend/src/pedigree.tsx
```diff
@@ -243,3 +259,121 @@ export function herdPedigree(genebanks: Genebank[], herdId: string | undefined,
...
+      }
+    }
+  }
+  nodes.push(offspringNode)
+
+  parents.forEach((parent: Individual) => {
+    edges.push({id: `${offspringNode.id}-${parent.number}`,
+                                  from: offspringNode.id,
+                                  to: parent.number})
+    //FEEDBACK, should it be optional to use calcPedigreeDan as well?
```
norling, line 349: we should put in the time to make a better calcPedigree function - none of the current ones should matter (imho). :)

frontend/src/pedigree.tsx
```diff
@@ -243,3 +259,121 @@ export function herdPedigree(genebanks: Genebank[], herdId: string | undefined,
...
+  edges = unique(edges, 'id')
+
+  // color nodes that are common ancestors
+  nodes.forEach(x => {
+    if (commonAnc.has(x['id'])) {
+      x.color.border = "#f24d0c"
+    }
+  })
+
+  //FEEDBACK, better to make a function that take edges as input and returns edges with the relevant edges colored?
```
norling: it might be worth the time to look over all the pedigree-views and functions tbh. I think we've all learned a bunch since this was written, so I think we could do a better overall design! We should talk to jonas about time though.

norling (approved):
Looks very neat in general! :) There are some minor things that I'd do differently, but it seems to work as intended!

frontend/src/inbreeding_form.tsx
inghylt: `setMale(newValue ? newValue : undefined)`

I wrote it like that because newValue is either Indivial or null. It became messier in other parts of the code with chosen male or female with type `Individual | null` in comparison with type `Individual | undefined`. But the solution felt a bit silly as it is only to satisfy typing complaints and not any logic :')

frontend/src/pedigree.tsx
inghylt: function unique and function commonAncestors are very similar at the moment. Was thinking that function unique could be rewritten to return both the unique elements but also which elements were duplicate instead of having two separate methods

norling: That seems like a good idea! less code is easier to maintain :)

jhagberg (changes requested):
No need to calculate inbreeding using the R-api the values are already calculated and available. For all registered and active animals. just get the intersection from the Mean kinship matrix. It gets a little tricker for grandparents but should be easily calculated using the values for the grandparents.

jhagberg (changes requested):
Looks like the kinship matrix is not chached on the python side
Adding for example
@CACHE.cached(timeout=300, key_prefix='R-api')
def get_kinship(g_id):

speeds up the calculations

pontus (approved):

pontus:
Looks good. I still am not sold on having as much logic in the frontend (particularly not without robust testing so it's more difficult for me to break stuff without noticing), but have no specific objections to merging this (I think getting this functionality is more important).

jhagberg:
Think @inghylt want to ad some more stuff or if we should do issues of whats left? I want some more changes before we merge but I can comment later today

frontend/src/testbreed_form.tsx
```diff
@@ -0,0 +1,320 @@
...
+}
+
+const emptyLabels: ancestorLabels = {
+  maleGFLabel: '',
+  maleGMLabel: '',
+  femaleGFLabel: '',
+  femaleGMLabel: ''
+}
+
+/**
```
jhagberg, line 97: Here they want to extend the range of individuals to test mate. It should not be only the active animals but also all added animals unregistered and registered for the 2 last years . Not sure how to do that maybe @norling has an idea?

frontend/src/testbreed_form.tsx
```diff
@@ -0,0 +1,320 @@
...
+                            />}
+                  />
+                  <div className={style.lineBreak}></div>
+                  <Autocomplete className={style.inputAncestor}
+                          disabled={individuals.female ? true : false}
+                          options={activeFemalesLimited}
+                          getOptionLabel={(option: LimitedIndividual) => individualLabel(option)}
+                          getOptionSelected={(option, value) => option.id === value.id}
+                          value={individuals.femaleGM}
+                          // FEEDBACK, is there a possibility for individuals to not have a registered mother/father?
```
jhagberg, line 202: No only founders should have missing parents. Otherwise its error in input

frontend/src/testbreed_recommendation.tsx
```diff
@@ -0,0 +1,199 @@
...
+      },
+      error => {
+      console.error(error);
+      userMessage(error, 'error')
+      } )
+    }, [chosenAncestors])
+
+  //TODO, error handling
+  let sufficientGenerations = true
+  //TODO, change depending on genebankId
```
jhagberg: Yes this should be
Gotlandskanin under 4%
Mellerudskanin under 15%

frontend/src/testbreed_recommendation.tsx
```diff
@@ -0,0 +1,199 @@
...
+  let father = chosenAncestors['male'] ? individualLabel(chosenAncestors['male']) : 
+  `♂(${chosenAncestors['maleGM'].name}+${chosenAncestors['maleGF'].name})`
+
+  let recommendationText
+  let recommendationSymbol
+  if (!sufficientGenerations) {
+    recommendationText = 'För få generationer tillgängliga för att göra tillförlitlig beräkning av inavelskoefficient'
+    recommendationSymbol = <CancelSharpIcon style={{ color: '#F44304', paddingRight: '4px'}} />
+  } else if (beneficialCOI) {
+    recommendationText = 'Parning rekommenderas'
```
jhagberg:
```suggestion
    recommendationText = 'Föreningen rekommenderar parningen'
```

frontend/src/testbreed_recommendation.tsx
```diff
@@ -0,0 +1,199 @@
...
+    recommendationSymbol = <CheckCircleSharpIcon  style={{ color: '#4CB950', paddingRight: '4px' }}/>
+  } else {
+    recommendationText = 'Parning rekommenderas ej'
```
jhagberg:
```suggestion
    recommendationText = 'Föreningen rekommenderar EJ parningen'
```

jhagberg (changes requested):

frontend/src/testbreed_form.tsx
pontus: I don't think I understand what this means "added animals unregistered" - that sounds like individuals with no known lineage, which I think means one can't say anything useful (please explain why I'm wrong)?

For the other part, I don't think there should be any problem, anything in the database should be fine I think.

jhagberg: I mean animals that not yet have a certificate. They exist but are not in the active population. 
for example, if a herd registers a mating event then this should create x number of individuals but they are not yet "registered" since tfhey do not have a certificate. They are not yet added into the active population but if other herds can find these animals and test-mate with them it can make it much easier to find new breeding pairs. O

We should also extend so animals that were active two years ago also should be findable in this list. 

Or if we can find animals in the big list and click "select for test-mate" like a compare computer thingy you can find in any webshop.

frontend/src/testbreed_recommendation.tsx
```diff
@@ -0,0 +1,213 @@
...
+      },
+      error => {
+        console.error(error)
+        userMessage('Something went wrong', 'error')
+      } )
+    }, [chosenAncestors])
+
+  //TODO, error handling
+  let calculationError = false
+  let thresholdCOI = genebankId === 1 ? 4.5 : 15
```
jhagberg:
```suggestion
  let thresholdCOI = genebankId === 1 ? 4 : 15
```

jhagberg: This should be 4 %

jhagberg (changes requested):
I cant test-mate melleruds kaniner I get "key error"

```
offspring_coi = kinship_matrix[payload["male"]][payload["female"]]
herdbook-main | KeyError: 'M1073-2031'
```

jhagberg (approved):
Not it also works! Great job!

PR #228, opened 2021-03-05
Update filter labels to match previously updated filter functionality

### LINKED ISSUE #246: Wrong filters in herd table
If the checkboxes for "Dölj döda" (hide dead) and "Dölj inaktiva djur" (hide inactive animals) are checked, more animals are shown than with unchecked boxes 

- Check if it's just the wrong way round (then exchange "hide"/"show") or if more is broken

inghylt:
Small fix. Filter functionality previously updated to show individuals when a filter is active and not hide them. The related labels were however only updated for src/genebank_view.tsx and not src/herd_view.tsx


jhagberg (approved):
LGTM not tested

pontus (approved):

PR #229, opened 2021-03-10
(Probably) fix issue #224

### LINKED ISSUE #224: Rabbit exists in two herds
There is a Data load bug if a rabbit is sold to an external herd the same year as it is born.
Then it will exist in two herds. 
I simple workaround is to change the line in the external herd tracking function. But could probably be solved in a better way

```
-               MAKE_DATE(DATE_PART('year', ht.herd_tracking_date)::integer, 12, 31)
+               MAKE_DATE(DATE_PART('year', ht.herd_tracking_date)::integer+1, 12, 31)

```

andkaha:
As suggested by @jhagberg, incrementing the year of transfer to an
external owner, may solve the issue of having an animal registered in
both a Gotlandskanin-herd and as "external" at the same time.

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -289,12 +289,13 @@ done | psql --quiet
 # Handle individuals that have disappeared from this genebank (animals
 # sold to external herds).  For each individual with "ny G" equal to
 # "0", figure out the most recent tracking date.  Add tracking info to
-# the "GX1" herd at the last of December of the year of that date.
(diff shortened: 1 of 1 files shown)


jhagberg (approved):
Yes this works for now. But dead rabbits should not belong to the GX1 herd if they got a dead note the same as the GX1 herd date. If we have a herd number in the year before column. 
But I asked Lekander how we should handle this so we can do this for now

PR #232, opened 2021-03-15
Fix missing birth_date in breeding table

### LINKED ISSUE #231: Founders is not listed in the Genebank list
Founders now have breeding ID but the breeding.father_id and mother_id is Null 

since the Individual.select in data_access.py is doing a lef join on Father.id == to Breeding.father_id the founders with Null is not selected I think.

andkaha:
Two changes:

1. Move the "Initial herd tracking" bit down after "Create empty
breeding events" as the latter actually changes the breeding table which
is part of creating the herd_tracking table.

2. Add UPDATE that moves birth dates in from the *_data tables when it's
missing in breeding.

Both changes applied to both `load-gotland.sh` and `load-mellerud.sh`.

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -165,7 +165,7 @@ psql --echo-errors --quiet <<-'END_SQL'
           d."Född" birth_date,
           MAX(d."Kull") litter_size
           FROM g_data d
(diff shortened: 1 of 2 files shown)


pontus (approved):
Unverified, but looks correct.

PR #233, opened 2021-03-18
Delete old loading script, dump full database at end of loading (but …

andkaha:
Tiny change in how the loading script dumps the database at the end.  Now it will dump the complete database, apart form the CSV tables.

I'm assuming that the `dump_load_db.sh` will be used for _actual_ dumping and restoring of the database, so that the plain text database dump created by the loading script can be used for debugging rather than to load from.

I have also deleted my old `load.sh` script with this PR.


pontus (approved):

jhagberg (approved):
ser la bra ut!

PR #234, opened 2021-03-18
Add test for duplicate numbers

### LINKED ISSUE #225: Import script handle duplicates
Now the import script just outputs an error if we have a duplicate. Would be good if it could be handled in a more human readable way

`ERROR:  duplicate key value violates unique constraint "individual_number_origin_herd_id"`

andkaha:
Check for duplicate numbers for both Gotland and Mellerud data sets.
If a duplicate number is found, issue a diagnostic message about this
and don't even attempt to load the data.

This should fix issue #225.

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -2,7 +2,14 @@
 
 # Needs to be called by load.sh
 
-echo "Loading rabbits"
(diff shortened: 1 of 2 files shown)


pontus (approved):
Unverified, looks correct.

andkaha:
This PR was forgotten about. Now merged.

PR #240, opened 2021-04-17
Bump jinja2 from known bad version

pontus:
Address https://github.com/NBISweden/theherdbook/security/dependabot/app/requirements.txt/jinja2/open


andkaha (approved):
This is probably ok.

PR #242, opened 2021-04-19
Various

andkaha:
Due to this morning's chatter around loading databases and setting up the project, I had a look at our database dump/restore script.

Changes:

* Rename `path` function `realpath` to overload the real `realpath` utility.
* Make `realpath` return pathnames relative to the current directory.
* Call `bash` using `env` (and set the `errexit` option with `set -e` instead).
* Create `help` function for help text output.
* Don't `cd` to script's directory (would cause unexpected path issue).
* Move output-related variables to "dump" section of code.
* Allow `--clean` option both before and after filename argument.
* Use `printf` for variable data output.
* Use array instead of string for `FLAGS`.
* Rearrange `if`-`then`-`else` body in `case ... esac` structure instead.
* Add `help` sub-command.
* Various cosmetic changes.

### DIFF
--- a/scripts/dump_load_db.sh
+++ b/scripts/dump_load_db.sh
@@ -1,92 +1,113 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
 
-cd $(dirname $0)
+set -e
 
-# some config variables
-function path {
-  if command -v realpath >/dev/null
-  then
-    realpath $1
-  else
-    echo "$1"
-  fi
+scriptdir=$( dirname "$0" )
+
+ENV_FILE="$scriptdir/../.docker/database-variables.env"
+CONTAINER='herdbook-db'
+
+realpath () {
+	if command -v realpath >/dev/null; then
+		command realpath --relative-to="$PWD" "$1"
+	else
+		printf '%s\n' "$1"
+	fi
 }
 
-OUTPUT_DIR="$(path ${2:-../db_dumps})"
-BASE="db-dump"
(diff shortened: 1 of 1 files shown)


scripts/dump_load_db.sh
```diff
@@ -1,92 +1,113 @@
...
-                                argument the dump file will be placed there, else
-                                in projectroot/db_dumps/
+	         help
+	                                Show this help text.
+	HELP
 
-HELP
-  exit 0
-fi
+	exit 1
```
pontus: There is an `exit 1` after calling for incorrect usage (and when provided because it's asked for, it makes more sense to exit with 0).

andkaha: Actually, since I put the help text into its own function, I think it makes more sense to exit from the caller instead than from the function.  Ok?  This `exit` is more or less a forgotten thing I was supposed to remove anyway.

pontus (approved):
Unverified, can't see anything to complain about.

(I'm not overly fond of arrays, but here I'm not fond of the alternative either.)

andkaha:
> Unverified, can't see anything to complain about.
> 
> (I'm not overly fond of arrays, but here I'm not fond of the alternative either.)

Well, making `FLAGS` an array is the only way to preserve the quoting of the expansion of the two variables in the options.  It's either using `FLAGS` as an array, duplicating the default options later in the code, or arranging the options to be part of the positional parameters and pick them out of there somehow.

PR #245, opened 2021-04-22
Make all linters happy

pontus:
Fixes to make the various linters in the super linter happy.


run_tests.sh
```diff
@@ -1,23 +1,25 @@
...
+cd "$(dirname "$0")" || exit 1
 
 if [ ! -d "$VENV" ]
 then
   echo "Warning: Could not find virtual environment '$VENV'."
   echo "Creating virtual environment '$VENV'".
 
-  python3 -m venv "$VENV"
-  [[ "$?" != "0" ]] && "Error: Couldn't create virtual environment" && exit 1
+  python3 -m venv "$VENV" || "Error: Couldn't create virtual environment" && exit 1
```
andkaha:
```
if ! python3 -m venv "$VENV"; then
    echo 'Error: Could not create virtual environment' >&2
    exit 1
fi
```

This would be unambiguous and also introduces the missing `echo`.

This is the only thing that _needs_ to be fixed.

scripts/load-in-docker.sh
```diff
@@ -69,6 +69,7 @@ for name in "$gfile" "$Gfile" "$mfile" "$Mfile"; do
 			csvname=${name%.xlsx}.csv
 
 			# Convert to CSV if CSV is missing or old
+			# shellcheck disable=SC2039
```
andkaha, line 72: This is for the `-ot`, right?

init_db.sh
```diff
@@ -26,6 +26,6 @@ echo 'Installing app dependencies...'
 pip3 install -r app/requirements.txt >/dev/null
 
 echo 'Initializing database...'
-cd app/
+cd app/ || exit 1
```
andkaha: You can drop the `/` in `app/` to make it similar to other similar statements in other scripts we have.

frontend/build.sh
```diff
@@ -7,13 +7,13 @@ esbuild                                          \
...
 mv main.js "$js_hash"
 css_hash="main.$(md5sum main.css | head -c 10).css"
 mv main.css "$css_hash"
-cat ../public/index.html |
-    sed 's,.*main.js.*,<script src="'"/$js_hash"'"></script>,' |
-    sed 's,<head>,&<link href="'"/$css_hash"'" rel="stylesheet">,' |
-    cat > index.html
+
+sed -e 's,.*main.js.*,<script src="'"/$js_hash"'"></script>,' \
+    -e 's,<head>,&<link href="'"/$css_hash"'" rel="stylesheet">,' \
```
andkaha, line 17: It's a _tiny_ bit odd that the `/` in front of the shell variable expansions aren't part of the single-quoted `sed -e` string, but instead in the double quoted string, i.e. that it is `'..."'"/$var"'"...'` instead of `'..."/'"$var"'"...'`.

andkaha (changes requested):
There is a missing `echo` that needs to be fixed. I have left other comments, but this is the only one that actually requires action.  That particular piece of code should also be fixed from a
```
command1 || command2 && command3
```
thing into a proper
```
if ! command1; then
    command2
    command3
fi
```

scripts/load-in-docker.sh
pontus: Yes.

andkaha (approved):
The things that I can comment on looks ok.

jonandernovella (approved):

PR #248, opened 2021-04-26
Change method for updating inds

jonandernovella:
This method manages updates of individuals, not users.


HenrikeW (approved):
Makes total sense!

pontus (approved):

PR #251, opened 2021-04-28
Add support for issuing certificates

jonandernovella:
- New endpoint for issuing certs under `/api/certificates/issue/in_id`
- PDF signatures created using the endesive library, which implements a CAdES-B handler
- Some fields have placeholder data and do not make sense yet


app/herdbook.py
```diff
@@ -474,6 +478,137 @@ def get_mean_kinship(g_id):
...
+    """
+    user_id = session.get("user_id", None)
+    ind = da.get_individual(i_number, user_id)
+
+    if ind:
+        all_data = get_all_data(ind, user_id)
+        pdf_bytes = get_certificate(all_data)
+        date = datetime.datetime.utcnow()
+        date = date.strftime("%Y-%m-%d_%H%M%S%f")
+        object_name = f"{date}.pdf"
```
pontus: Can we generate a more usable file name?

jonandernovella: I guess so! We need to decide on a pattern for that. I am open tu suggestions.

app/herdbook.py
```diff
@@ -474,6 +478,137 @@ def get_mean_kinship(g_id):
...
+
+def _get_parent(ind, user_id, ancestry_level, ancestry_type):
+    ancestries = {
+        (1, "F"): "father",
+        (1, "M"): "mother",
+        (2, "F"): "father",
+        (2, "MF"): "father",
+        (2, "FM"): "mother",
+        (2, "MM"): "mother",
+    }
```
pontus: Is this used elsewhere? Otherwise it's seems to complicate stuff rather than simplifying to me.

jonandernovella: Complicate what stuff? And no, I don't think it's used elsewhere.

pontus:
Constant file names - move these to `utils.settings` (pick up from environment but fine to have default as current)?

pontus:
A lot of names contain pkcs, if anything that should probably be x509, but I don't see why one would need that.

app/herdbook.py
jhagberg: The number of the rabbit is welknowed I suggest number-certificate.pdf

So like G1274-2001-5678966.pdf

Maybe date but that's probably to much

app/herdbook.py
pontus: I haven't tried rewritring it, but my spontaneous feeling is it would be simpler having it as a single function.

app/herdbook.py
```diff
@@ -474,6 +480,302 @@ def get_mean_kinship(g_id):
...
+            uploaded = upload_certificate(
+                pdf_bytes=signed_data.getvalue(), ind_number=ind["number"]
+            )
+        except Exception as ex:  # pylint: disable=broad-except
+            APP.logger.info("Unexpected error while updating certificate "+str(ex))
+            return jsonify({"response": "Error processing your request"}), 400
+
+        if uploaded:
+            return create_pdf_response(
+                pdf_bytes=signed_data, obj_name=f'{ind["certificate"]}.pdf'
```
```suggestion
                pdf_bytes=signed_data, obj_name=f'{ind["digital_certificate"]}.pdf'
```

app/herdbook.py
```diff
@@ -474,6 +480,304 @@ def get_mean_kinship(g_id):
...
+    if ind is None:
+        return jsonify({"response": "Individual not found"}), 404
+    certificate_exists = ind.get("certificate", None)
+    if certificate_exists:
+        return jsonify({"response": "Certificate already exists"}), 400
+
+    ##cert_number = str(random.randint(50001, 99999))
+    cert_number = "99999"
+    ind.update(**request.json, certificate=cert_number)
+    da.update_individual(ind, session.get("user_id", None))
```
pontus, line 537: I don't understand how this can work as is, and also think it's not the right way to do it.

```suggestion
    ind.certificate = cert_number
    ind.save()
```

? 

I've seen https://docs.peewee-orm.com/en/latest/peewee/querying.html#atomic-updates but do not think it applies here.

PR #253, opened 2021-05-04
Load 2021 Gotland data, and "number" health-check.

### LINKED ISSUE #252: Healthcheck: Year of birth vs. number
@jhagberg suggested adding a health check that find animals whose "number" indicates that they've been born in a year which is different from the birth year recorded.

andkaha:
1. Minor documentation tweaks.
2. Also load Gotland data for 2021.
3. Correct bug that prevented Gotland 2019's body fat data from being loaded if available.
4. Add health-check that makes sure that the individual's "number" corresponds to the birth year.


jhagberg (approved):
LGTM works with the latest data file.

PR #254, opened 2021-05-04
Add support for certificate uploads

jonandernovella:
- Issuing and updating endpoints can now upload the pdfs to S3.
- boto3 library used for interacting with S3
- minio dev server is used

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -590,14 +590,6 @@ def upload_certificate(pdf_bytes, prefix, certificate_number):
     )
 
 
-def check_certificate_version_exists(certificate_number, checksum):
-    """
-    Returns a boolean value specifying if a certificate version already exists in S3.
-    """
-    cert_checksums = s3.get_s3_client().list_object_checksums(prefix=certificate_number)
-    return checksum in cert_checksums
-
-
 def check_certificate_s3(object_name):
     """
     Returns a boolean value specifying if any certificate already exists in S3.
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -586,7 +586,7 @@ def upload_certificate(pdf_bytes, ind_number):
     Triggers a S3 certificate upload
     """
     return s3.get_s3_client().put_object(
-        bucket_file_name=f"{ind_number}.pdf", file_data=pdf_bytes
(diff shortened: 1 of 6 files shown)


app/herdbook.py
```diff
@@ -513,14 +523,22 @@ def issue_certificate(i_number):
     if certificate_exists:
         return jsonify({"response": "Certificate already exists"}), 400
 
-    ind.update(**request.form)
+    ## TODO: Make sure this is unique
+    cert_number = str(random.randint(50001, 99999))
```
pontus, line 527: There is a todo but these would normally be a series, so we should keep track and issue the next.

app/utils/s3.py
```diff
@@ -0,0 +1,128 @@
...
+        secret_key,
+        access_key,
+        verify,
+        cert,
+        private_key,
+    ):
+        """
+        Constructor for initialising a S3 client.
+        """
+        assert isinstance(bucket, str)
```
pontus, line 29: I'm personally not overall fond of this kind - if something can fulfill the role it should be okay, and alterantively if one wants stronger typing, I prefer using `typing` and static tools to analyse.

app/utils/s3.py
```diff
@@ -0,0 +1,128 @@
...
+            endpoint_url=self.endpoint,
+            region_name=region,
+            use_ssl=self.endpoint.startswith("https"),
+            verify=verify,
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+            config=config,
+        )
+
+        print(vars(self))
```
pontus, line 65: These are easy to forget, better to write a suitable log statement directly?

app/herdbook.py
jonandernovella: This will be fixed in a future PR.

app/utils/s3.py
jonandernovella: I am not either. But it felt like a quick stopper for undesired values passed into the configuration. We could try out [Mypy](https://github.com/python/mypy) in the next iteration?

app/utils/s3.py
jonandernovella: Yes. Good that you found this one. We can set it as a trace or debug line.

PR #255, opened 2021-05-05
Upgrade some js packages

pontus:
Bump versions of most packages to what works (e.g.`@date-io` is held back due to compability issues).

Fix so frontend runs with `node_modules` from the volume from image creation, not whatever happens to be in the `frontend` directory of the developer.


jonandernovella (approved):
Great work!

PR #257, opened 2021-05-05
Support certificate verification

jonandernovella:
- Endpoint for verification under `/api/certificates/verify/ind_number`
- Checksum and signature verification is performed


app/herdbook.py
```diff
@@ -497,13 +497,19 @@ def update_certificate(i_number):
...
-           pdf_bytes=pdf_bytes.getvalue(), ind_number=ind["number"]
-        )
+
+        try:
+            signed_data = sign_data(pdf_bytes)
+            uploaded = upload_certificate(
+                pdf_bytes=signed_data.getvalue(), ind_number=ind["number"]
+            )
+        except Exception as ex:  # pylint: disable=broad-except
+            print(ex)
```
```suggestion
            APP.logger.info("Unexpected error while updating certificate "+str(ex))
```

app/herdbook.py
```diff
@@ -523,20 +529,26 @@ def issue_certificate(i_number):
...
+    try:
+        signed_data = sign_data(pdf_bytes)
+        uploaded = upload_certificate(
+            pdf_bytes=signed_data.getvalue(), ind_number=ind_number
+        )
+    except Exception as ex:  # pylint: disable=broad-except
+        print(ex)
+        return jsonify({"response": "Error processing your request"}), 400
 
     if uploaded:
```
pontus, line 586: I'm not overly fond of this style because one needs to think about the flow to understand why this doesn't blow up with `uploaded` undefined. I much prefer either having the return within the try or making sure to define `uploaded` to `None` above the try.

app/herdbook.py
```diff
@@ -558,6 +570,38 @@ def preview_certificate(i_number):
...
+        return jsonify({"response": "Individual not found"}), 404
+
+    uploaded_bytes = request.get_data()
+
+    try:
+        checksum = hashlib.sha256(uploaded_bytes).hexdigest()
+        signed = verify_signature(uploaded_bytes)
+        present = verify_certificate_checksum(ind["number"], checksum=checksum)
+    except Exception as ex:  # pylint: disable=broad-except
+        print(ex)
```
```suggestion
        APP.logger.info("Unexpected error while verifying certificate "+str(ex))
```

app/herdbook.py
```diff
@@ -558,6 +570,38 @@ def preview_certificate(i_number):
...
+        return jsonify({"response": "Error processing your request"}), 400
+
+    if present and signed:
```
pontus, line 637: Same as above with preference for doing within try or defining before.

pontus (approved):

app/herdbook.py
jonandernovella: This one is for capturing mostly S3 related failures, which are a bit tricky to propagate otherwise. I am happy to work on a better solution in the next iteration :)

app/herdbook.py
jonandernovella: Yes, that's good. For now I will make them all debug lines though.

app/herdbook.py
pontus: I think I may have been unclear, I'd be happy if there would be an `uploaded = False` above the code.

As far as I can see, the code above works (won't blow up on `uploaded` undefined), but it isn't obvious.

jonandernovella: Aha! Alright now I follow you.

app/herdbook.py
jonandernovella: Done

app/herdbook.py
jonandernovella: Should be fixed now

PR #258, opened 2021-05-05
Functionality to issue a new digital certificate

HenrikeW:
The dialogue that lets the user issue a certificate is rendered in individual_certificate.txt. That's where all the new functionality can be found, except from the button that opens the dialogue in the first place. It is found in individual_view.txt.

A lot of the form and parts of the functionality are copied from the individual_edit-form. To reduce iterative code, the form could instead get its own component, but in that case, I think some local states should be handled globally. So for now, I left this for future improvements.

During developing I installed Prettier which auto-formats the code on save. Thus, some files that I only briefly touched can have seemingly large changes that mostly consist of added semicolons, for example.


pontus:
Apparently  the linter misses a few commits (I'm not sure why but guess this is due to some merging), the super-linter can be run locally with e.g.

```
docker run -e RUN_LOCAL=true -v $PWD/:/tmp/lint ghcr.io/github/super-linter:latest
```

app/herdbook.py
```diff
@@ -474,6 +480,250 @@ def get_mean_kinship(g_id):
...
+    Returns an updated pdf of the individual given by `i_number`.
+    """
+    user_id = session.get("user_id", None)
+    ind = da.get_individual(i_number, user_id)
+
+    if ind is None:
+        return jsonify({"response": "Individual not found"}), 404
+
+    certificate_exists = ind.get("certificate", None)
+    if certificate_exists and check_certificate_s3(object_name=ind["number"]):
```
```suggestion
    if certificate_exists and check_certificate_s3(object_name="certificate/" + ind["number"]):
```

Alternatively `ind["number"] + "/certificate"`, feels like a good idea to keep the generic identifier free if we possibly  want to add other storage objects.

pontus (approved):

frontend/src/individual_certificate.tsx
```diff
@@ -0,0 +1,614 @@
...
+  const [previewUrl, setPreviewUrl] = React.useState(undefined as unknown);
+  const [certificateUrl, setCertificateUrl] = React.useState("");
+  //States to make pdf-preview-library work
+  const [numPages, setNumPages] = React.useState(null);
+  const [pageNumber, setPageNumber] = React.useState(1);
+
+  const { user } = useUserContext();
+  const { genebanks, colors } = useDataContext();
+  const { popup } = useMessageContext();
+  const { userMessage } = useMessageContext();
```
inghylt, line 140: Minor detail, but one can do `const {userMessage, popup} = useMessageContext()`

frontend/src/individual_certificate.tsx
```diff
@@ -0,0 +1,614 @@
...
+  };
+
+  /**
+   * Fetch individual data from the backend
+   */
+  React.useEffect(() => {
+    user && user.canEdit(id)
+      ? get(`/api/individual/${id}`).then(
+          (data: Individual) => {
+            console.log(data);
```
inghylt, line 168: There several console.log in this file. I guess it's intentional?

frontend/src/individual_certificate.tsx
```diff
@@ -0,0 +1,614 @@
...
+        }
+      })
+      .catch((error) => {
+        {
+          userMessage(error.message, "error");
+        }
+      });
+  };
+
+  // This function is necessary to make the previw work.
```
inghylt, line 266: Minor typo

inghylt (commented):
Good work! Made some minor comments. One thought, individual_certificate.tsx is quite a long file. Might be worth trying to split things up in separate files if that makes sense logically

PR #259, opened 2021-05-06
Feature/new certs frontend

HenrikeW:
New version of my former PR.


pontus:
Created because of my confused comments, this work was done and merged in #258.

PR #265, opened 2021-05-06
All files in /src formatted with Prettier extension

HenrikeW:
In total, 30 files were auto-formatted using the VSCode extension Prettier. All files are .tsx-files.


pontus (approved):
Seems legit

PR #266, opened 2021-05-06
Add cache as suggested and start job refreshing cache at startup

pontus:
Use flask-cache to keep kinship values cached.

Use apscheduler to add job to get kinship often, triggering cache load if needed.

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -415,7 +415,6 @@ def get_inbreeding(g_id):
     return {}
 
 
-@CACHE.cached(timeout=KINSHIP_LIFETIME, key_prefix="R-api")
 @APP.route("/api/<int:g_id>/kinship/")
 def kinship(g_id):
     """
@@ -423,7 +422,7 @@ def kinship(g_id):
     """
     return jsonify(get_kinship(str(g_id)))
 
-
+@CACHE.cached(timeout=KINSHIP_LIFETIME, key_prefix="R-api")
 def get_kinship(g_id):
     """
     Fetch kinship matrix from R-api of the genebank given  by `g_id`.
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -422,6 +422,7 @@ def kinship(g_id):
     """
     return jsonify(get_kinship(str(g_id)))
 
+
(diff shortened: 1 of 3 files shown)


app/herdbook.py
```diff
@@ -49,6 +50,8 @@
 LOGIN = LoginManager(APP)
 LOGIN.login_view = "/login"
 
+KINSHIP_LIFETIME = 300
```
jhagberg, line 53: This could probably be longer. Very long if we can force to update the cache when adding new animals?

app/herdbook.py
```diff
@@ -412,6 +415,7 @@ def get_inbreeding(g_id):
     return {}
 
 
+@CACHE.cached(timeout=KINSHIP_LIFETIME, key_prefix="R-api")
```
jhagberg: the CACHE should be on the internal  get_kinship(p.id)

app/herdbook.py
jhagberg: Alright, so you are updating the kinshipmatrix every 15 s that will work but will reload with the same data many times. could we instead force a cache to reload when new animals get registered? But this could probably work but if we check every 30 min and cache is 1 hour long or something. They usually update this matrix once per year.

inghylt (approved):
Nice!

jhagberg (changes requested):

app/herdbook.py
pontus: No, from what I understand (and my interpretations of what I've seen), data is not reloaded every time, instead the incurred cost is the cost of getting the cached data every 15 seconds.

The first calls before the `r-api` is up and answering will fail, so having a low value is a "practical" work around to trying to sync that.

We can certainly increase the cache lifetime and add invalidation on updates.

app/herdbook.py
pontus: Ah, I must have messed things up when pushing vs testing, true and would give pessimal behaviour.

pontus: Fixed.

app/herdbook.py
jhagberg: Okey so its realoded ever 300 s ?

jhagberg (approved):

jhagberg:
Lets merge this for now

PR #267, opened 2021-05-07
Fix logic for user selector

### LINKED ISSUE #244: Can't edit users with the user editor
Creating new users seems to work as expected (and one can choose permissions and so on), but editing users does not seem to work.

pontus:
Fixes logic to let user editor understand the selected user.

Fixes #244.

The user editor uses URL parts as `/manage/user/1` where 1 would be the selected user. Make sure to accept the user from that.


HenrikeW (approved):
Seems fair.

PR #270, opened 2021-05-10
Change colour to color, with migration

pontus:
Use "color" instead of "colour".

Also do migration of old database.


jhagberg (approved):
LGTM not tested but cant see any issues with this

jonandernovella (approved):

PR #271, opened 2021-05-11
Issue certs with unique ids

### LINKED ISSUE #268: Issue certs with unique id numbers
These should be within the 100000-Inf range and preferrably persisted in the database. 

` cert_number = str(random.randint(100000, Inf))`

DoD: Whenever a new digital certificate is issued, the individual will be assigned an immutable and unique certificate number.

jonandernovella:
- Used a SQL sequence for auto generation of unique ids within a series.
- Official certificate numbers become the digital id when issuing a digital certificate.

Closes #268

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -532,7 +532,10 @@ def issue_certificate(i_number):
     if certificate_exists:
         return jsonify({"response": "Certificate already exists"}), 400
 
-    ind.update(**request.json, certificate=cert_number)
+    form = request.json
+    cert_number = ind["digital_certificate"]
+
+    ind.update(**form, certificate=cert_number)
     da.update_individual(ind, session.get("user_id", None))
 
     data = get_certificate_data(ind, user_id)
@@ -569,7 +572,7 @@ def preview_certificate(i_number):
     form = request.json
 
     data = get_certificate_data(ind, user_id)
-    data.update(**form)
+    data.update(**form, certificate=ind["digital_certificate"])
     pdf_bytes = get_certificate(data)
     return create_pdf_response(pdf_bytes=pdf_bytes, obj_name="preview.pdf")
 
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -569,13 +569,13 @@ def preview_certificate(i_number):
     if ind is None:
         return jsonify({"response": "Individual not found"}), 404
 
+    data = get_certificate_data(ind, user_id)
+
     if request.method == "POST":
         form = request.json
-        data = get_certificate_data(ind, user_id)
         data.update(**form, certificate=ind["digital_certificate"])
         pdf_bytes = get_certificate(data)
     elif request.method == "GET":
-        data = get_certificate_data(ind, user_id)
         pdf_bytes = get_certificate(data)
 
     return create_pdf_response(pdf_bytes=pdf_bytes, obj_name="preview.pdf")
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -558,7 +558,7 @@ def issue_certificate(i_number):
     return jsonify({"response": "Certificate could not be uploaded"}), 400
 
 
-@APP.route("/api/certificates/preview/<i_number>", methods=["POST"])
+@APP.route("/api/certificates/preview/<i_number>", methods=["POST", "GET"])
 @login_required
 def preview_certificate(i_number):
     """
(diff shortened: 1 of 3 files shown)


app/utils/database.py
```diff
@@ -398,6 +398,8 @@ class Meta:  # pylint: disable=too-few-public-methods
 
         indexes = ((("mother", "father", "birth_date"), True),)
 
+## Create sequence to allow unique ids for digital certificates
+DATABASE.execute_sql('''CREATE SEQUENCE IF NOT EXISTS certificates_seq START WITH 100000 INCREMENT BY 1 MAXVALUE 199999 NO CYCLE''')
```
pontus: Move this to `init`?

I think sqlite doesn't support `CREATE SEQUENCE`, so only run if using postgres. 

(How does this match with tests?)

app/utils/database.py
```diff
@@ -410,6 +412,7 @@ class Individual(BaseModel):
     origin_herd = ForeignKeyField(Herd)
     name = CharField(50, null=True)
     certificate = CharField(20, null=True)
+    digital_certificate = IntegerField(sequence="certificates_seq", unique=True)
```
pontus: Allow `null` (e.g. with `null=True`)?

pontus:
Migration?

pontus:
Maybe I'm not following along as I need to - does "Official certificate numbers become the digital id when issuing a digital certificate." mean that we copy the value into `certificate`?

If not - can we not separate individuals without certificate from those who have one from the database only?

app/utils/database.py
jonandernovella: done

jonandernovella:
> Maybe I'm not following along as I need to - does "Official certificate numbers become the digital id when issuing a digital certificate." mean that we copy the value into `certificate`?
> 
> If not - can we not separate individuals without certificate from those who have one from the database only?

Yes, we do copy the value into certificate.

app/utils/database.py
jonandernovella: It should be now in init(). I am not sure what to do about sqlite.

pontus:
> > Maybe I'm not following along as I need to - does "Official certificate numbers become the digital id when issuing a digital certificate." mean that we copy the value into `certificate`?
> > If not - can we not separate individuals without certificate from those who have one from the database only?
> 
> Yes, we do copy the value into certificate.

Ok, good, do you recall where this happens (I don't see it quickly)?

app/herdbook.py
```diff
@@ -531,9 +532,10 @@ def issue_certificate(i_number):
...
         return jsonify({"response": "Certificate already exists"}), 400
 
-    ##cert_number = str(random.randint(50001, 99999))
-    cert_number = "99999"
-    ind.update(**request.json, certificate=cert_number)
+    form = request.json
+    cert_number = ind["digital_certificate"]
+
+    ind.update(**form, certificate=cert_number)
     da.update_individual(ind, session.get("user_id", None))
```
pontus: Same as https://github.com/NBISweden/theherdbook/pull/251/files#r635138222

app/utils/database.py
```diff
@@ -410,6 +409,7 @@ class Individual(BaseModel):
     origin_herd = ForeignKeyField(Herd)
     name = CharField(50, null=True)
     certificate = CharField(20, null=True)
+    digital_certificate = IntegerField(sequence="certificates_seq", unique=True, null=True)
```
pontus, line 412: I think this needs to be fixed to work with test as well, either by moving test to postgres or supporting sqlite, but I don't know exactly how yet.

I don't see that it needs to delay this specific PR, though (but should be fixed before merge to develop).

app/utils/database.py
```diff
@@ -1091,6 +1091,9 @@ def init():
     with DATABASE.atomic():
         sh_bootstrap.save()
 
+    ## Create sequence to allow unique ids for digital certificates
+    DATABASE.execute_sql('''CREATE SEQUENCE IF NOT EXISTS certificates_seq START WITH 100000 INCREMENT BY 1 MAXVALUE 199999 NO CYCLE''')
```
pontus:
```suggestion
    if type(DATABASE) == PostgresqlDatabase:
        DATABASE.execute_sql('''CREATE SEQUENCE IF NOT EXISTS certificates_seq START WITH 100000 INCREMENT BY 1 MAXVALUE 199999 NO CYCLE''')
```

pontus (approved):

PR #276, opened 2021-05-18
Adding frontend for updating a certificate (incl. bugfix for issue and preview certificates)

HenrikeW:
Two minor bugs were fixed:
- adding a content-type to the headers of the fetch request in `previewCertificate` 
- adding content-type, a return statement and more specific error handling in the fetch request in issueCertificate

Moreover, the user interface to update a certificate was added.

### DIFF
--- a/frontend/src/individual_certificate.tsx
+++ b/frontend/src/individual_certificate.tsx
@@ -300,7 +300,6 @@ export function IndividualCertificate({
         }
       })
       .then((blob) => {
-        console.log("blob", blob);
         if (blob) {
           setCertificateUrl(window.URL.createObjectURL(blob));
           setShowSummary(false);
--- a/frontend/src/individual_certificate.tsx
+++ b/frontend/src/individual_certificate.tsx
@@ -233,7 +233,7 @@ export function IndividualCertificate({
       });
   };
 
-  // Returns the signed certificate.
+  // Returns the signed, new certificate.
   const issueCertificate = (id: string, content: any) => {
     fetch(`/api/certificates/issue/${id}`, {
       body: JSON.stringify(content),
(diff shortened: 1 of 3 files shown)


frontend/src/individual_certificate.tsx
```diff
@@ -244,10 +251,55 @@ export function IndividualCertificate({ id }: { id: string }) {
...
+          throw new Error("Något gick fel.");
+        }
+      })
+      .then((blob) => {
+        if (blob) {
+          setCertificateUrl(window.URL.createObjectURL(blob));
+          setShowSummary(false);
+          setShowComplete(true);
+        } else {
+          throw new Error("Något gick fel.");
```
pontus:
```suggestion
          throw new Error("Något gick fel (det här borde inte hända).");
```

?

This is one problematic area - for translation work, it's helpful to use the same string when possible. At the same time, having differing messages is useful for getting more helpful bug reports at times.

frontend/src/individual_certificate.tsx
```diff
@@ -244,10 +251,55 @@ export function IndividualCertificate({ id }: { id: string }) {
...
+      body: JSON.stringify(content),
+      method: "PATCH",
+      credentials: "same-origin",
+      headers: {
+        Accept: "application/pdf",
+        "Content-Type": "application/json",
+      },
+    })
+      .then((res) => {
+        if (res.status === 400) {
```
pontus: Does 400 have a specific meaning here that's essentially different from the other error, otherwise move up the other cases and let the default handle this?

frontend/src/individual_certificate.tsx
```diff
@@ -115,7 +115,13 @@ const useStyles = makeStyles({
   },
 });
 
-export function IndividualCertificate({ id }: { id: string }) {
+export function IndividualCertificate({
+  id,
+  action,
+}: {
+  id: string;
+  action: string;
```
pontus, line 123: I have a slight preference for using the language strengths and going for an `enum` here.

frontend/src/individual_certificate.tsx
```diff
@@ -585,26 +637,40 @@ export function IndividualCertificate({ id }: { id: string }) {
               color="primary"
               disabled={isUserGood ? false : true}
               onClick={() => {
-                issueCertificate(id, certificateData);
+                action == "issue"
```
pontus: This is probably fine (and short, which is nice), but I have a slight preference for e.g. `switch` as it gives me a cuddly feeling of more clearly implying multivalue and thus less risk of missing changes required if adding new `action` values.

But I don't have that much typescript experience, so the best case would be to use (non-confusing) language features if they can help enforce/flag things like this.

pontus (approved):
:+1: 

Looks good.

frontend/src/individual_certificate.tsx
HenrikeW: yes true, there is no specific case for 400 (any more). Removed it as you suggested.

frontend/src/individual_certificate.tsx
HenrikeW: good point! changed it.

jonandernovella (approved):

PR #277, opened 2021-05-18
Use memoize rather than cache for function

pontus:
Bugfix, `cached` with `key_prefix` collapsed the argument internally.


jhagberg (approved):
LGTM yes of course we should use memoize that also saves the arguments. Good catch!

PR #281, opened 2021-05-20
Colour is color now

pontus:
Fix import scripts to the current database names.


jhagberg (approved):
LGTM tested in production :)

PR #285, opened 2021-05-24
Add tests for certificate endpoints

jonandernovella:
- Preview, issue, update and verify are now tested for correct bahaviour
- More tests can be added if considered neccessary
- TODO: test for the upcoming endpoint for uploading certificates not issued by the system


HenrikeW (approved):
Looks good!

PR #289, opened 2021-05-26
User interface for verifying certificates

HenrikeW:
In certificate_verification:
- a component to upload a file and send it to verify_certificates endpoint
- three different, designed messages for valid, outdated and invalid certificates
- for outdated certificates option to send a request to preview_certificate to see a preview of the version stored in S3
- a simple general error message for all other cases

In individual_view:
A button that will open the verification-component in a popup

**If** you have time/possibility: please switch to my branch and try verifying some files. (If you want a valid / outdated one, you need to issue/ update a cert first, obviously.) I've had problems with some of the endpoints lately, and I don't think it's related to this code, but I'd like to be sure.


frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+          } else if (res.status === 202) {
+            setCertOutdated(true);
+          } else if (res.status === 404) {
+            setCertNotFound(true);
+          } else {
+            throw new Error("Något gick fel.");
+          }
+        })
+        .catch((error) => {
+          userMessage(error.message, "error");
```
jonandernovella, line 103: What does this do?

jonandernovella: Checking `useMessageContext`.

jonandernovella: I see this will print a dialog on screen. Is this correct?

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+    })
+      .then((res) => res.arrayBuffer())
+      .then((data) => {
+        setPreviewWidth(window.innerWidth * 0.5);
+        setPreviewUrl(data);
+      })
+      .catch((error) => {
+        userMessage(error.message, "error");
+      });
+  };
```
jonandernovella, line 136: Maybe this code should be in `communication.tsx`together with all the other HTTP request handlers.

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+            <CheckCircle className={style.successIcon} />
+          </div>
+          <p>
+            Dokumentet innehåller den senaste informationen om{" "}
+            {individual?.name} och finns i vår databas.
+          </p>
+        </Box>
+      ) : certOutdated ? (
+        <Box
+          border={3}
```
jonandernovella, line 201: why are these within brackets?

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+          color="primary"
+          onClick={() => {
+            setCertValid(false);
+            setCertOutdated(false);
+            setCertNotFound(false);
+            setShowPdf(false);
+            verifyCertificate(id);
+          }}
+          disabled={file ? false : true}
+        >
```
jonandernovella, line 179: What about allowing the user deleting the file before it's sent to verify? is this important?

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+          </p>
+          <Button
+            variant="outlined"
+            color="primary"
+            onClick={() => {
+              setShowPdf(true);
+              getLatestCertificate(id);
+            }}
+          >
+            Se senaste certifikat
```
jonandernovella, line 228: What about "See preview of the latest certificate"?

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+      ) : (
+        <div></div>
+      )}
+      {showPdf ? (
+        <Document
+          file={previewUrl}
+          onLoadSuccess={onDocumentLoadSuccess}
+          renderAnnotationLayer={true}
+          loading={<CircularProgress />}
+          noData=""
```
jonandernovella, line 257: What does `onData` do?

jonandernovella (approved):

jonandernovella:
Tested and verified that works as intended!

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+          userMessage(error.message, "error");
+        });
+    };
+
+    if (file) {
+      const reader = new FileReader();
+      reader.onload = (event) => {
+        const binaryFile = event.target?.result
+          ? event.target.result
+          : undefined;
```
pontus, line 112: Notify user of error instead?

frontend/src/certificate_verification.tsx
pontus, line 201:
```suggestion
          border=3
```

As mentioned by Jon.

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+          </div>
+          <p>
+            Dokumentet innehåller den senaste informationen om{" "}
+            {individual?.name} och finns i vår databas.
+          </p>
+        </Box>
+      ) : certOutdated ? (
+        <Box
+          border={3}
+          borderRadius={8}
```
pontus, line 202:
```suggestion
          borderRadius=8
```

pontus (approved):
Looks good.

frontend/src/certificate_verification.tsx
pontus: For clarity; Notification if I remember correctly (dialog implies some form of persistence, whereas these will go away if I remember correctly).

frontend/src/certificate_verification.tsx
pontus: This shouldn't be different from other failures to read the file, as handled in the `onload` event.

frontend/src/certificate_verification.tsx
```diff
@@ -0,0 +1,266 @@
...
+            throw new Error("Något gick fel.");
+          }
+        })
+        .catch((error) => {
+          userMessage(error.message, "error");
+        });
+    };
+
+    if (file) {
+      const reader = new FileReader();
```
pontus, line 108: I can't see it being an issue here, but if I understand this correctly (which might not be the case), it will first load the file and then post it. If possible, streaming it would be preferential.

(E.g. what happens if I pick a 50 Gbyte file rather than the one I intended.)

pontus: (I think API support used to be iffy, but I have not kept up to date and wouldn't know about anything resembling modern things.)

PR #291, opened 2021-05-27
Try to make python linters happy

jonandernovella:
Small cleanup to comply with Black's expectations.


app/utils/certificates.py
```diff
@@ -57,7 +57,8 @@
     "MormorFärg": "MM_color",
     "Nummer": "number",
     "IntygasAv": "user_id",  # Username
-    "IntygasNummer": "certificate",  # Herd id of rabbit individual (Without herd identifier) G/M
+    # Herd id of rabbit individual (Without herd identifier) G/M
+    "IntygasNummer": "certificate",
```
HenrikeW, line 61: What was the problem here? Is it because the comment was to long to be in the same line?

app/utils/database.py
```diff
@@ -1100,8 +1099,9 @@ def init():
     with DATABASE.atomic():
         sh_bootstrap.save()
 
-    ## Create sequence to allow unique ids for digital certificates
-    if type(DATABASE) == PostgresqlDatabase:
+    # Create sequence to allow unique ids for digital certificates
+    if isinstance(DATABASE, PostgresqlDatabase):
+        # pylint: disable=E501
```
HenrikeW: Out of curiosity: What does this pylint-comment do? What is being disabled?

app/utils/s3.py
```diff
@@ -23,7 +24,7 @@ def __init__(  # pylint: disable=too-many-arguments
         verify,
         cert,
         private_key,
-    ):
+    ):  # pylint: disable=too-many-arguments
```
HenrikeW, line 27: even here I'm curious: why would pylint complain about too many arguments, and why do we want to have many arguments here anyway?

app/utils/database.py
```diff
@@ -1100,10 +1099,11 @@ def init():
     with DATABASE.atomic():
         sh_bootstrap.save()
 
-    ## Create sequence to allow unique ids for digital certificates
-    if type(DATABASE) == PostgresqlDatabase:
+    # Create sequence to allow unique ids for digital certificates
+    if isinstance(DATABASE, PostgresqlDatabase):
+        # pylint: disable=C0301
```
HenrikeW, line 1104: Out of curiosity: what is it we are disabling here?

frontend/src/individual_certificate.tsx
```diff
@@ -274,6 +274,7 @@ export function IndividualCertificate({
   };
 
   // Returns the updated certificate.
+  // jscpd:ignore-start
```
HenrikeW, line 277: what does this line do?

scripts/load-docker-compose.sh
```diff
@@ -13,6 +13,7 @@
 #	          -M herd-registry-mellerud.xlsx
 #
 
+# shellcheck disable=SC2266
```
HenrikeW, line 16: and what are we disabling here?

scripts/load-in-docker.sh
```diff
@@ -69,7 +69,7 @@ for name in "$gfile" "$Gfile" "$mfile" "$Mfile"; do
 			csvname=${name%.xlsx}.csv
 
 			# Convert to CSV if CSV is missing or old
-			# shellcheck disable=SC2039
+			# shellcheck disable=SC2039,SC3013
```
HenrikeW, line 72: and here?

HenrikeW (approved):
The linters seem to be happy now.

scripts/load-docker-compose.sh
pontus: We pipe to a `test`-like expression (`[[ .. ]]`) which is typically not correct.

scripts/load-in-docker.sh
pontus: `-ot` is not a POSIX thing, but we still want to use it.

app/utils/database.py
pontus: Long line (almost all linter disablers can be found with the corresponding linter and code, e.g. `pylint C0301` will have a page explaining this as the first hit).

app/utils/database.py
pontus: This was likely meant as a flake8 disablement (for long lines).

app/utils/s3.py
pontus: Having too many arguments can mean there's a higher risk of switching up the order (and it's more difficult to spot). I'm not sure how pylint considers it, but one common solution for python is to use keyword arguments instead.

The "requested" arguments are those that are part of the "complete" specification of a s3 resource, so they'll be needed to support a generic s3 resource.

pontus (approved):
:+1:

PR #294, opened 2021-05-28
Add support for certificate operations

jonandernovella:
- Individual certificates can be issued, updated, previewed and verified
- They are stored in a S3 backend
- Tests work with a mock S3 lib


pontus:
Test database shouldn't be checked in. Can we get a script to generate test keys/certificates rather than having them checked in?

frontend/src/individual_certificate.tsx
```diff
@@ -370,26 +420,46 @@ export function IndividualCertificate({ id }: { id: string }) {
...
+          {action == "issue" ? (
+            <h1>Certifikatet är klart!</h1>
+          ) : (
+            <h1>Certifikatet uppdaterades!</h1>
+          )}
+          <div className={style.paneControls}>
+            <a
+              target="_blank"
+              href={certificateUrl}
+              download={individual.number}
```
pontus: Do we really need `target="_blank"` here? I have not yet succeeded in testing, but my understanding is that the `download` tag should trigger a file download in anything resembling modern (I'm yet unable to understand if `download` and `target="_blank"` will give a flashing window that disappear, an empty window or something else).

app/herdbook.py
```diff
@@ -505,6 +509,314 @@ def testbreed():
...
+
+    certificate_exists = ind.get("certificate", None)
+    form = request.json
+    uploaded = False
+
+    try:
+        present = check_certificate_s3(ind_number=ind["number"])
+        if certificate_exists and present:
+            data = get_certificate_data(ind, user_id)
+            data.update(**form)
```
pontus: I still don't like this and don't think it does what's intended (at least not always).

app/herdbook.py
```diff
@@ -505,6 +509,314 @@ def testbreed():
...
+            pdf_bytes = get_certificate(data)
+            signed_data = sign_data(pdf_bytes)
+            uploaded = upload_certificate(
+                pdf_bytes=signed_data.getvalue(), ind_number=ind["number"]
+            )
+            ind.update(**form)
+            da.update_individual(ind, session.get("user_id", None))
+    except Exception as ex:  # pylint: disable=broad-except
+        APP.logger.info("Unexpected error while updating certificate " + str(ex))
```
pontus, line 541: I think the "can't update paper certificate" case deserves enough treatment that the frontend can give a sensible error.

app/herdbook.py
```diff
@@ -505,6 +509,314 @@ def testbreed():
...
+    if ind is None:
+        return jsonify({"response": "Individual not found"}), 404
+    certificate_exists = ind.get("certificate", None)
+    if certificate_exists:
+        return jsonify({"response": "Certificate already exists"}), 400
+
+    form = request.json
+    cert_number = ind["digital_certificate"]
+
+    ind.update(**form, certificate=cert_number)
```
pontus: Same for database update here as before.

app/herdbook.py
```diff
@@ -505,6 +509,314 @@ def testbreed():
...
+    # version 2 means size 42x42
+    qr_x_len, qr_y_len = 42, 42
+
+    certificate = certs.CertificateGenerator(
+        form=settings.certs.template,
+        form_keys=certs.FORM_KEYS,
+    )
+
+    qr_code = certs.QRHandler(
+        link=settings.service.host + "/individual/" + data["number"] + "/verify",
```
pontus, line 800: I'm fine with merging this for now, but think the host should really be picked up from the request (likely requires adding 
`proxy_set_header Host $host;` in nginx).

app/herdbook.py
jonandernovella: I still don't understand what you want either. Can you please explain?

frontend/src/individual_certificate.tsx
jonandernovella: @HenrikeW Will this be addressed here or in the other PR?

frontend/src/individual_view.tsx
```diff
@@ -208,9 +208,23 @@ export function IndividualView({ id }: { id: string }) {
...
+              {user?.canEdit(id) && (
+                <Button
+                  className={style.editButton}
+                  variant="contained"
+                  color="primary"
+                  onClick={() =>
+                    popup(<IndividualCertificate id={id} action={"update"} />)
+                  }
+                >
+                  Uppdatera certifikat
```
pontus: After some playing and thinking, I believe we need to do this nicer, even for a MVP.

If there is a paper certificate these should not be shown at all as far as I understand (digital certificate can neither be created nor updated).

If there is a digital certificate "Uppdatera certifikat" should be available.

If there is no certificate, "Skapa nytt certifikat" should be shown.

If I understand the current design correctly, this should be possibly by testing if integer conversion of `certificate` equals integer conversion of `digital_certificate`.

app/herdbook.py
jonandernovella: The HTTP status code(404) returned by the exception handler is mostly for when calling the API with an individual that does not have a digital certificate uploaded **or the backend is broken**. Hopefully there is a paper one but this is not guaranteed I think this will mostly happen when the user tries to update the certificate of **the wrong individual,** though. I am fine with changing this but note it will not always convey the correct information.

frontend/src/individual_view.tsx
jonandernovella: In the design stage we contemplated these options but later decided not to do this to simplify things.

app/herdbook.py
pontus: Hrrm, misplaced this comment, should be on the `ind.update` line.

For that; what it looks like it should do is start creating an update query (http://docs.peewee-orm.com/en/latest/peewee/api.html#Model.update which doesn't change anything). Then call the `update_individual` to do update (which is intended to update from a form which is not exactly what we're doing here).

So again, I don't understand how the ind.update can change anything (given as for my current understanding, it just creates an update query that is not executed but immediately forgotten). 

And if it does work as intended, I still don't like it on account on being confusing.

I'd much prefer having a whitelist of allowed changes here (keys in dict `form`) and passing to a new function in `data_access` to update individual according to that form. That could either do a

```
    ind = get_individual(i_number, user_id)
    ind.update(**form).where(Individual.number == i_number).execute()
```   
or
```
    ind = get_individual(i_number, user_id)
    for k,v in form:
        ind.setattr(k,v)
    ind.save()
```   

(both inside `with DATABASE.atomic():`)

app/herdbook.py
pontus: Ok, after some thinking I believe we really should improve the frontend anyway and this won't be such a big thing.

app/herdbook.py
pontus: See https://github.com/NBISweden/theherdbook/commit/ef1bfbe4aa910ba6e6d7fe4ac650fda8db325ab8  for some minumum changes (only naming and comments) to lessen confusion. I'd prefer other changes but that would bloat this PR even more, so it's something for the future.

pontus (approved):

PR #296, opened 2021-05-31
add noopener/noreferrer to download links for cert

HenrikeW:
to make GitHub Code Scanning happier

### DIFF
--- a/frontend/src/individual_certificate.tsx
+++ b/frontend/src/individual_certificate.tsx
@@ -446,7 +446,6 @@ export function IndividualCertificate({
           )}
           <div className={style.paneControls}>
             <a
-              target="_blank"
               href={certificateUrl}
               download={individual.number}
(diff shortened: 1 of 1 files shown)


frontend/src/individual_certificate.tsx
```diff
@@ -446,15 +446,15 @@ export function IndividualCertificate({
           )}
           <div className={style.paneControls}>
             <a
-              target="_blank"
```
jonandernovella, line 449: Why did this disappear?

jonandernovella (approved):

frontend/src/individual_certificate.tsx
pontus: Not sure if we need to both delete it and add `rel` (but don't think it hurts).

Since this should be a download, why should there be a new window?

pontus (approved):

PR #298, opened 2021-06-01
frontend for verification process

HenrikeW:
including some fixes:
- buttons now show depending on if there is a certificate
- if file upload doesn't work, the user gets an error message
- the messages to the user during the verification process are now correctly representing the different cases


jonandernovella (approved):
Great work!

PR #300, opened 2021-06-02
Minor fix of documentation for R service use

pontus:


jonandernovella (approved):

jonandernovella (changes requested):
Thr arguments of the register user script are user, pass and then email

jonandernovella (approved):

PR #302, opened 2021-06-03
Fix some fields of the certificates

jonandernovella:
Specifically genebank numbers and herd identifiers have been updated.


jhagberg (approved):
Think this looks right, Can we have the User name instead of the ID ?

jonandernovella:
DO NOT MERGE

PR #307, opened 2021-06-07
Ensure cert numbers start from 100K

### LINKED ISSUE #268: Issue certs with unique id numbers

jonandernovella:
This has been done by only updating the difital certificate id upon certificate issuing.

### DIFF
--- a/app/tests/test_endpoints.py
+++ b/app/tests/test_endpoints.py
@@ -315,6 +315,12 @@ class TestEndpoints(FlaskTest):
             self.assertEqual(pdf_response.headers["Content-Type"], "application/pdf")
             self.assertEqual(pdf_response.status_code, 200)
 
+            # Issue a certificate again
+            response = context.post(
+                f"/api/certificates/issue/{individual_2}", json=valid_issue_form
+            )
+            self.assertEqual(response.status_code, 400)
+
             # Download a certificate
             download_response = context.get(f"/api/certificates/issue/{individual}")
             self.assertEqual(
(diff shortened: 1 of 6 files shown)


app/tests/test_endpoints.py
```diff
@@ -294,15 +295,26 @@ def test_certificate_ops(self):
...
                 f"/api/certificates/issue/{individual}", json=valid_issue_form
             )
-            certificate = db.Individual.get(self.individuals[0].id).certificate
-            self.assertEqual(certificate, "G1310-2121")
-            color = db.Individual.get(self.individuals[0].id).color.id
-            self.assertEqual(color, 3)
+            digital_certificate = db.Individual.get(
+                self.individuals[0].id
+            ).digital_certificate
+            self.assertEqual(digital_certificate, 1000000)
```
jhagberg: should start at 100 000 , 100K  not 1M

jhagberg:
```suggestion
            self.assertEqual(digital_certificate, 100000)
```

app/utils/data_access.py
```diff
@@ -663,7 +663,14 @@ def update_individual(form, user_uuid):
         form["herd"] = form["herd"].get("herd", None)
     if not Herd.select().where(Herd.herd == form["herd"]).exists():
         return {"status": "error", "message": "Individual must have a valid herd"}
-
+    if form.get("issue_digital", False):
+        nextval = 1000000
```
jhagberg: same here I think  should be 100K

jhagberg (changes requested):
Should start at 100K

jhagberg (approved):
LGTM

PR #311, opened 2021-06-10
making the testbreeding form prettier

HenrikeW:
Two changes:

1. (code structure): moving the function activeIndividuals from the file testbreed_form to data_context_global to make it reusable in the add-rabbits-functionality
2. (code formatting): testbreed_form is now formatted with the "prettier" extension for more consistency in the frontend part of the codebase.

So most changes are formatting changes (semicolons, double/single quotes etc)

Issue #305, opened 2021-06-07
Sensibility checks for registering rabbits


pontus:
We need to define some sensibility checks, e.g. aliveness at some point.

PR #312, opened 2021-06-10
Users can now add rabbits 

HenrikeW:
User interface to add a new rabbit to a herd. 
A herd owner can add a rabbit to her own herd. Managers and admins can add to every herd. 

Error handling still needs to be improved (let the backend send proper HTTP error codes). 
Also there are no checks for new individuals, you can e.g. easily create two with the same number. Backend fix?
(See also issue #305).

Review / merge PR #311 first to avoid the marked changes being spoiled by a lot of formatting changes.


jonandernovella:
Working on the data validation part

HenrikeW:
I'm not sure what about the failing linter tests. I don't think what is reported there is a real problem. At least I don't see how to write it shorter (and clearer!) as it is now.

pontus:
jscpd can be a bit overzealous at times, but I'm still very fond of having the super-linter do all it does without configuring tons of things.

Obviously jscpd thinks we should refactor but I haven't so far. If you search there are loads of examples of disabling and enabling jscpd.

HenrikeW:
Thanks, Pontus! I also think it's good to have the linter, just disabled it for a couple of lines. 
I now implemented error handling for the case when a new individual's number already exists. (Rebased on `develop` which already includes Jon's backend solution for this). So I think this can now be merged.

PR #314, opened 2021-06-11
Add fullname for users

pontus:
Add full name (`fullname`) for users, will be needed for e.g. certificates.


jonandernovella (approved):
Great!

app/utils/database.py
```diff
@@ -1270,12 +1280,41 @@ def migrate_5_to_6():
...
-            version=5,
+            version=6,
             comment="Remove password_hash from hbuser",
             applied=datetime.now(),
         ).execute()
 
 
+def migrate_6_to_7():
+    """
+    Migrate between schema version 5 and 6.
```
HenrikeW, line 1291: probably a wrong comment here, with regard to the function's name?

app/utils/data_access.py
```diff
@@ -439,7 +445,7 @@ def update_user(form, user_uuid=None):
         return {"status": "error", "message": f"malformed request: {form}"}
 
     # Check permissions - allow users to update e-mail only
-    if not (user.is_admin or user.is_manager) or (
+    if not (user.is_admin or user.is_manager) and (
```
HenrikeW, line 448: I don't really understand why this needed to be changed.

HenrikeW (approved):

app/utils/database.py
pontus: Yes, :eyes: .

PR #315, opened 2021-06-11
Replace user id with fullname in certificates

jonandernovella:
Certificate now contains full name of logged in user


HenrikeW (approved):
Looks good!

PR #317, opened 2021-06-14
add new rabbits to a herd (more tidy PR)

HenrikeW:
This should contain all the code needed to add a rabbit to a herd that is not on `develop`yet. 

A herd owner can add rabbits to her own herd, managers and admins can add rabbits to every herd. 

It's not possible any more to add a rabbit with an already existing number.


pontus (approved):
Not verified, but looks good!

HenrikeW:
do you want another review, Pontus? Or merge the PR? Or do you want me to do that?

PR #318, opened 2021-06-14
Display all data from pedigree correctly

jonandernovella:
The certificates did not display data from all individuals because tthe individual object was mutated in the update data access method

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -777,6 +777,7 @@ def issue_certificate(i_number):
         cert_number = res.get("digital_certificate", None)
 
         cert_data = get_certificate_data(ind_data_copy, user_id)
+        cert_data.update(**form, certificate=cert_number)
         pdf_bytes = get_certificate(cert_data)
         ind_number = ind_data["number"]
         uploaded = False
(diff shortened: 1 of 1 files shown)


app/herdbook.py
```diff
@@ -770,12 +771,12 @@ def issue_certificate(i_number):
         form = request.json
 
         ind_data.update(**form, issue_digital=True)
+        # keep the ind_data object intact
+        ind_data_copy = copy.copy(ind_data)
         res = da.update_individual(ind_data, session.get("user_id", None))
```
pontus, line 776: Is `update_individual` equipped to handle these dicts with Individuals in them?

pontus (approved):

app/herdbook.py
jonandernovella: Nothing has changed in that regard. I think it still expects a form, and that's what it currently gets.

pontus: Ok, I guess nothing has changed at least (if I read correctly, what it gets is one of those dicts with Individuals in them).

PR #320, opened 2021-06-14
Get fullname from Google if available

pontus:
Also; user should be logged in directly if account is created.


jonandernovella (approved):
Great!

PR #323, opened 2021-06-16
Track newly added individuals in their herd

jonandernovella:
Newly created individuals now are tracked within their current herd. This is done by adding a new entry to the herd tracking table.

We also have a new field called selling_date which is sent in the form optionally if the rabbit is directly sold and registered in a different herd.

### DIFF
--- a/app/utils/data_access.py
+++ b/app/utils/data_access.py
@@ -876,7 +876,6 @@ def add_individual(form, user_uuid):
 
     individual.save()
 
-    setattr(individual, "herdtracking_set", True)
     update_herdtracking_values(
         individual=individual,
         herd=individual.origin_herd,
@@ -886,14 +885,13 @@ def add_individual(form, user_uuid):
 
 
 def update_herdtracking_values(individual, herd, user_signature):
-    if individual.herdtracking_set:
-        HerdTracking(
-            from_herd=herd,
-            herd=individual.origin_herd,
-            signature=user_signature,
-            individual=individual,
-            herd_tracking_date=datetime.now(),
-        ).save()
+    HerdTracking(
+        from_herd=herd,
+        herd=individual.origin_herd,
+        signature=user_signature,
+        individual=individual,
+        herd_tracking_date=datetime.now(),
+    ).save()
 
 
 def update_individual(form, user_uuid):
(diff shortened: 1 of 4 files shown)


app/utils/data_access.py
```diff
@@ -874,12 +874,28 @@ def add_individual(form, user_uuid):
     if "bodyfat" in form:
         update_bodyfat(individual, form["bodyfat"])
 
-    # TODO: also create herd tracking values
-
     individual.save()
+
+    setattr(individual, "herdtracking_set", True)
```
pontus: I don't think I follow the logic of adding this - if one don't want to update herd tracking, why not just not call it?

pontus (approved):
Seems workable, although I have some questions about how it's done (the `setattr`).

app/utils/data_access.py
jonandernovella: I assumed that this was the case. But we can make it a flag of course.

pontus:
```suggestion
```

app/utils/data_access.py
```diff
@@ -874,12 +874,28 @@ def add_individual(form, user_uuid):
...
 
+def update_herdtracking_values(individual, herd, user_signature):
+    if individual.herdtracking_set:
+        HerdTracking(
+            from_herd=herd,
+            herd=individual.origin_herd,
+            signature=user_signature,
+            individual=individual,
+            herd_tracking_date=datetime.now(),
+        ).save()
```
pontus:
```suggestion
   HerdTracking(
        from_herd=herd,
        herd=individual.origin_herd,
        signature=user_signature,
        individual=individual,
        herd_tracking_date=datetime.now(),
    ).save()
```

app/utils/data_access.py
```diff
@@ -874,12 +874,28 @@ def add_individual(form, user_uuid):
...
 
 
+def update_herdtracking_values(individual, herd, user_signature):
+    if individual.herdtracking_set:
+        HerdTracking(
+            from_herd=herd,
+            herd=individual.origin_herd,
+            signature=user_signature,
+            individual=individual,
+            herd_tracking_date=datetime.now(),
```
jhagberg: The first date should be the same date as the animal is born on and the founding herd

pontus:
The from herd should also really be a parameter (and null for newly added rabbits).

PR #324, opened 2021-06-17
Return breeding id from register breeding endpoint

jonandernovella:
The register breeding event endpoint now returns the breeding id. This makes possible for our service to register birth events.


pontus (approved):
Untested, looks good!

PR #325, opened 2021-06-17
updates for the add-rabbits feature

HenrikeW:
Major changes:
- added Register to top navigation for genebank managers. This way, they don't need to navigate the herds to add an individual
- form now has fields for the current herd and a selling date
- if an individual has been successfully created, the user can choose between two options: add another rabbit (clears all fields) or add a sibling (leaves pre-filled what is identical between the siblings)

Still missing: 
so far, no breeding event is created


jonandernovella (approved):
LGTM

frontend/src/individual_add.tsx
```diff
@@ -94,125 +163,252 @@ export function IndividualAdd({ id }: { id: string }) {
...
+    setHerdKey(Date.now());
+    setColorKey(Date.now());
+    setIndividual({} as Individual);
+    setSuccess(false);
+  };
+
+  const resetSibling = () => {
+    const numberParts: string[] = individual?.number?.split("-");
+    const sibling: Individual = {
+      number: numberParts ? numberParts[0] + "-" : null,
```
pontus, line 266: I don't think I understand this, do we not end up with just the herd part here?

pontus (approved):
Looks good!

jonandernovella (approved):

pontus:
Some additional notes; 

I don't consider myself very proficient in typescript/javascript, so I like following advice. One common such tends to be to prefer strict equals (`===` rather than `==`).

The other note general observation is that I prefer to "fail early" - essentially to reduce conditional depth by handling any failure cases specifically early on and return for those cases. To me, it makes things easier to understand because I have less to keep track of.

Some language designers agree and have that as part of recommended how you use the language, for others it "just style". It's totally fine disagreeing, but I thought I should mention that.

PR #327, opened 2021-06-18
add function to create breeding event

HenrikeW:
When clicking on "Skapa", 
- a breeding event is created,
- it's updated with birth information and
- the individual is created.
Three API-requests chained.

The newly created individual will have the field "breeding" which connects it with information about its parents and litter size.

### DIFF
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -263,14 +263,15 @@ export function IndividualAdd({
 
     // if there is no match, create a new breeding
     if (!breedingMatch) {
-      newBreeding = await createBreeding(currentBreeding);
+      const newBreeding: { breeding_id: number; status: string } =
+        await createBreeding(currentBreeding);
 
       if (!newBreeding) {
         // createBreeding will show a message, so we should not try to
         return;
       }
 
-      birth = { ...birth, id: newBreeding };
+      birth = { ...birth, id: newBreeding.breeding_id };
     } else {
       birth = { ...birth, id: breedingMatch.id };
     }
@@ -311,27 +312,37 @@ export function IndividualAdd({
     }
   };
 
-  const createBreeding = async (breedingData: Object): string | null => {
+  const createBreeding = async (
+    breedingData: Object
+  ): { status: string; message?: string; breeding_id?: number } => {
     const breedingEvent = await post("/api/breeding", breedingData);
 
     if (breedingEvent.status == "success") {
       return breedingEvent.breeding_id;
     }
 
-    const translate: Map<string, string> = {
-      "Not logged in": "Du är inte inloggad. Logga in och försök igen",
-      "Unknown mother":
+    const translate: Map<string, string> = new Map([
+      ["Not logged in", "Du är inte inloggad. Logga in och försök igen"],
+      [
+        "Unknown mother",
         "Okänd mor, modern måste vara en aktiv individ i databasen",
-      "Unknown father":
+      ],
+      [
+        "Unknown father",
         "Okänd far, fadern måste vara en aktiv individ i databasen",
-      Forbidden: "Du har inte behörighet att lägga till individen",
-    };
+      ],
+      [
+        "Unknown mother, Unknown father",
+        "Okända föräldrar. Både modern och fadern måste vara aktiva individer i databasen.",
+      ],
+      ["Forbidden", "Du har inte behörighet att lägga till individen"],
+    ]);
 
     if (
       breedingEvent.status == "error" &&
       translate.has(breedingEvent.message)
     ) {
-      userMessage(translate[breedingEvent.message], "error");
+      userMessage(translate.get(breedingEvent.message), "error");
       return;
     }
 
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -170,11 +170,11 @@ export function IndividualAdd({
     };
   });
 
-  // creates a breeding event, updates it with birth information and calls the createIndividual function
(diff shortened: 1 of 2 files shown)


frontend/src/individual_add.tsx
```diff
@@ -158,6 +159,51 @@ export function IndividualAdd({
...
+      };
+
+      let birth = {
+        date: individual.birth_date,
+        litter: individual.litter,
+        id: null,
+      };
+
+      const breedingEvent = await post("/api/breeding", breeding);
+
```
pontus, line 268: Error handling? 

And we should probably check if this breeding exist before trying to create it (as mentioned, I don't know if the API is there, but we can add if needed).

HenrikeW: Right. My thought was to have this one, pretty generic error handling for both requests:
```
 } else {
        userMessage(
          // register_birth in data_access.py requires a litter size.
          // Thus, the field must be mandatory in the form. Okay for föreningen?
          "Något gick fel. Kolla så att du har lagt till mor, far, kullstorlek och födelsedatum.",
          "error"
        );
      }
```
It will be triggered if something in one of the requests goes wrong, but maybe it's not so beautiful. I can add more specific error handling to the first request to prevent sending a second request that is meant to fail from the very beginning. 

Also, the backend checks if there already is a breeding, it looks like this in `register_breeding` in `data_access.py`: 
```
exists = (
        Breeding.select()
        .where(Breeding.father == father)
        .where(Breeding.mother == mother)
        .where(Breeding.breed_date == breed_date)
        .count()
    )
    if exists:
        return {"status": "error", "message": "Breeding already registered"}
```
But I would need the endpoint to send the breeding's id even in this case. For now, it's only send in case a new breeding is created. So that's something that needs to be added, and I don't think I can do that myself.

pontus: I don't see a problem with a generic error message, but if I understand things correctly, that relies on doing the second request and failing - is that incorrect?

(I would possibly expect to get some other error because of `breeding_id` not being present in `breedingEvent`, but I guess that doesn't happen.)

pontus: Regarding backend; if it simplifies I guess we can change the add breeding api endpoint to allow existing breeding to just return that. The alternative would be to get the breeding for the herd and check if it exists already.

frontend/src/individual_add.tsx
```diff
@@ -158,6 +159,106 @@ export function IndividualAdd({
...
+        breedingEvent.status == "error" &&
+        breedingEvent.message.includes("Unknown father" || "Unknown mother")
+      ) {
+        userMessage(
+          "En eller båda föräldrar kunde inte hittas. Både mor och far måste vara aktiva individer i databasen.",
+          "warning"
+        );
+      } else if (
+        breedingEvent.status == "error" &&
+        breedingEvent.massage == "Forbidden"
```
pontus:
```suggestion
        breedingEvent.message == "Forbidden"
```

pontus: But I'm sorry if that was unclear, I don't have a problem with generic messages. And if we want to offer better messages in the frontend, I think we should define error codes rather than rely on strings like htis.

frontend/src/individual_add.tsx
```diff
@@ -158,6 +159,106 @@ export function IndividualAdd({
...
+      ) {
+        userMessage(
+          "Du har inte behörighet att lägga till den här individen.",
+          "error"
+        );
+      }
+
```
pontus, line 279: It will still (currently) continue here, right?

frontend/src/individual_add.tsx
```diff
@@ -187,59 +200,97 @@ export function IndividualAdd({
...
+          };
+
+          // Get all the breedings for the individual's origin herd and see if there is a match
+          const breedings: { breedings: Breeding[] } = await get(
+            `/api/breeding/${originHerd.herd}`
           );
+          const breedingMatch = breedings.breedings.find((item) => {
+            item.mother == currentBreeding.mother &&
+              item.father == currentBreeding.father &&
+              item.breed_date == currentBreeding.date;
```
pontus:
```suggestion
              (item.breed_date == currentBreeding.date ||
               item.birth_date == individual.birth_date);
```

?

frontend/src/individual_add.tsx
```diff
@@ -187,59 +200,97 @@ export function IndividualAdd({
...
+              "Något gick fel på grund av tekniska problem. Kontakta en administrator.",
+              "error"
+            );
+          }
         } else {
-          userMessage("Individens nummer har inget giltigt format.", "warning");
+          userMessage(
+            "Första delen i individens nummer motsvarar ingen besättning.",
+            "warning"
+          );
```
pontus: I'm having a hard time keeping track of all the cases here, would it be possible to decrease the depth and have simple 
if cases that return instead?

pontus (approved):
Untested, but believe this will work.

frontend/src/individual_add.tsx
HenrikeW: So now we check if the breeding exists by getting all the breedings for the current herd. Only there is no breeding a request is sent to create one.

frontend/src/individual_add.tsx
HenrikeW: Yes, I agree! I'd also love to have some error codes instead of strings. It would be easier and also more in line with standards. Now even an error from the backend becomes a "200 OK" in the browser.

frontend/src/individual_add.tsx
HenrikeW: Good catch!

Issue #328, opened 2021-06-21
Multiple entries in herd_tracking gives duplicates in individuals api call


pontus:
R-api for hopefully current data shows in logs 

```
Duplicated IDs were removed.
This includes e.g.
         Indiv      Sire       Dam    Sex Born is_active
8636 G684-0911 G529-0732 G529-0741 female 2009     FALSE
9280 G684-0912 G529-0732 G529-0741 female 2009     FALSE
9282 G684-0913 G529-0732 G529-0741 female 2009     FALSE
```

I believe this may be related to

```
herdbook=# select * from herd_tracking where individual_id=8636;
 herd_tracking_id | from_herd_id | herd_id | signature_id | individual_id | herd_tracking_date 
------------------+--------------+---------+--------------+---------------+--------------------
             8636 |              |     406 |              |          8636 | 2008-06-30
            15860 |              |     406 |              |          8636 | 2008-12-31
            16483 |              |     406 |              |          8636 | 2009-12-31
            17136 |              |     406 |              |          8636 | 2010-12-31
(4 rows)

herdbook=# 
```

but am not certain; for the other individuals we have

```
herdbook=# select * from herd_tracking where individual_id=9280;
 herd_tracking_id | from_herd_id | herd_id | signature_id | individual_id | herd_tracking_date 
------------------+--------------+---------+--------------+---------------+--------------------
             9280 |              |     422 |              |          9280 | 2019-04-29
            27042 |              |     422 |              |          9280 | 2020-12-31
(2 rows)

herdbook=# select * from herd_tracking where individual_id=9282;
 herd_tracking_id | from_herd_id | herd_id | signature_id | individual_id | herd_tracking_date 
------------------+--------------+---------+--------------+---------------+--------------------
             9282 |              |     422 |              |          9282 | 2019-05-03
            25949 |              |      36 |              |          9282 | 2019-12-31
            27043 |              |      36 |              |          9282 | 2020-12-31
(3 rows)

herdbook=# 


```

Can we see what would cause the query at https://github.com/NBISweden/theherdbook/blob/develop/app/utils/data_access.py#L1074 to give the same individual multiple times?

pontus:
Upon further inspection I think the herd tracking may be a red herring, e.g. 8636 only shows up once as an individual in the API - but as male (and as father). So why does optisel conclude 8636 is female?

andkaha:
[...]

PR #330, opened 2021-06-22
Feature/issue 328

andkaha:
This should take care of resolving issue #328 by simply turning the first query in `get_individuals()` (against the `herd_tracking` table) into a `SELECT DISTINCT` query.


pontus (approved):
Verified working.

PR #332, opened 2021-06-23
let createBreeding return complete breedingEvent object

HenrikeW:
Feature didn't work. Discovered the bug. Fixed it.


pontus (approved):
Looks legit.

PR #333, opened 2021-06-24
Update herd tracking if needed on individual updates and fallout

pontus:
Use explicit queries to the HerdTracking table rather than
herdtracking_set. Test updates to reflect this.

Some additional cleanup, e.g. not using the original herd as from_herd
in all herd tracking updates.

### DIFF
--- a/app/tests/test_data_access.py
+++ b/app/tests/test_data_access.py
@@ -447,9 +447,6 @@ class TestDataAccess(DatabaseTest):
                 and ind.death_date is None
                 and not ind.death_note
             )
-            print("active is")
-            print(active)
-            print(ind.is_active)
             ind_info = {
                 "id": ind.id,
                 "name": ind.name,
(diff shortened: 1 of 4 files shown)


app/tests/test_data_access.py
```diff
@@ -452,6 +483,11 @@ def parent(data):
             gb0_expected += [ind_info]
 
         gb0_value = da.get_individuals(self.genebanks[0].id, self.admin.uuid)
+
+        print("hej")
+        print(gb0_value)
+        print(gb0_expected)
```
jonandernovella: !!

jonandernovella:
```suggestion
```

app/tests/test_data_access.py
```diff
@@ -407,27 +417,48 @@ def parent(data):
...
+                    .count()
+                    else False
+                )
                 and (ind.current_herd.is_active or ind.current_herd.is_active is None)
                 and ind.death_date is None
                 and not ind.death_note
             )
+            print("active is")
+            print(active)
+            print(ind.is_active)
```
jonandernovella:
```suggestion
```

jonandernovella (changes requested):

jonandernovella (approved):

Issue #334, opened 2021-06-28
user.canEdit function gives wrong permissions when id is an individual's id


HenrikeW:
_Situation_: `user_context.tsx` contains a function `canEdit` that checks if a user can edit a herd, individual or genebank by looking at the `id` of that individual, herd or genebank. 
If id is an individual's id, the function first generates a herd id by taking the first part of the individual-id and then checks if the user has permission to edit this herd. 
The function is used for instance in `individual_view.tsx` to check if the buttons "Redigera individ" and "Sälj individ" are shown or not.

_What's wrong?_ The first part of an individual's id is its **origin herd**. If the rabbit has been sold, not the owner of the origin herd but the owner of the individual's **current herd** should be able to edit it. 

_Idea for fixing:_ use the individual's `id` to make a query to `/api/individual`, look at the field `herd` in the response and then check if the user has permission to edit that herd.

PR #335, opened 2021-06-28
Feature sell rabbits

HenrikeW:
A part of the form used for adding new animals was turned into a reusable component and now even serves for selling individuals. 
The button for selling individuals in `individual_view` should only appear if you have the permission to sell this animal (i.e. if you are an admin, the genebank manager or the owner of the individual's current herd). 
There is a bug related to this described in issue #334 . 

I tested the feature manually with an admin user and it works fine. 
When I tried with a user who only owns a certain herd, I got a 502 response back and can't find the cause of that problem. 
I'd be happy if you could give it a try as well.

### DIFF
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -105,9 +105,6 @@ export function IndividualAdd({
   const [currentGenebank, setCurrentGenebank] = React.useState(
     genebank ? genebank : (undefined as Genebank | undefined)
   );
-  const [herdOptions, setHerdOptions] = React.useState(
-    genebank ? genebank.herds : ([] as LimitedHerd[])
-  );
   const [success, setSuccess] = React.useState(false as boolean);
   // states to handle the Autocompletes rerendering
   const [herdKey, setHerdKey] = React.useState(0 as number);
@@ -496,7 +493,7 @@ export function IndividualAdd({
               </h2>
               <IndividualSellingform
                 individual={individual}
-                herdOptions={herdOptions}
+                herdOptions={genebank ? genebank.herds : []}
                 herdKey={herdKey}
                 onUpdateIndividual={handleUpdateIndividual}
               />
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -28,7 +28,7 @@ import { useMessageContext } from "@app/message_context";
 import { get, post } from "@app/communication";
 import { useUserContext } from "./user_context";
 import { useDataContext } from "./data_context";
-import { IndividualSellingform } from "./individual_sellingform";
+import { IndividualSellingForm } from "./individual_sellingform";
 
 const useStyles = makeStyles({
   paneControls: {
@@ -491,7 +491,7 @@ export function IndividualAdd({
               <h2 className={style.sellingTitle}>
                 Fyll i bara om kaninen har sålts
               </h2>
-              <IndividualSellingform
+              <IndividualSellingForm
                 individual={individual}
                 herdOptions={genebank ? genebank.herds : []}
                 herdKey={herdKey}
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -28,6 +28,7 @@ import { useMessageContext } from "@app/message_context";
 import { get, post } from "@app/communication";
 import { useUserContext } from "./user_context";
 import { useDataContext } from "./data_context";
+import { IndividualSellingform } from "./individual_sellingform";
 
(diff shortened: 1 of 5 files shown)


pontus:
Would it be possible to remove/disable the sell button once there is a proper submitted sale? Once the sale goes through, a herd owner no longer has access to the animal and might be confused and try to sell it again.

frontend/src/individual_sellingform.tsx
```diff
@@ -0,0 +1,76 @@
...
+  LimitedHerd,
+} from "./data_context_global";
+
+const useStyles = makeStyles({
+  inputField: {
+    margin: "1em 0",
+  },
+});
+
+export function IndividualSellingform({
```
pontus:
```suggestion
export function IndividualSellingForm({
```

frontend/src/individual_add.tsx
```diff
@@ -28,6 +28,7 @@ import { useMessageContext } from "@app/message_context";
 import { get, post } from "@app/communication";
 import { useUserContext } from "./user_context";
 import { useDataContext } from "./data_context";
+import { IndividualSellingform } from "./individual_sellingform";
```
pontus:
```suggestion
import { IndividualSellingForm } from "./individual_sellingform";
```

frontend/src/individual_add.tsx
```diff
@@ -493,45 +491,12 @@ export function IndividualAdd({
...
-                    variant="outlined"
-                    margin="normal"
-                    helperText="Lämna tom om kaninen inte har sålts"
-                  />
-                )}
-                className={style.ancestorInput}
-                onChange={(event, newValue) => {
-                  newValue && handleUpdateIndividual("herd", newValue.herd);
-                }}
+              <IndividualSellingform
```
pontus:
```suggestion
              <IndividualSellingForm
```

frontend/src/individual_sellingform.tsx
```diff
@@ -0,0 +1,76 @@
...
+  individual,
+  herdOptions,
+  herdKey,
```
pontus, line 27: I'm not thrilled with these names. And what is `herdKey` here?

HenrikeW:
> Would it be possible to remove/disable the sell button once there is a proper submitted sale? Once the sale goes through, a herd owner no longer has access to the animal and might be confused and try to sell it again.

The button should only be visible if the individual you look at belongs to your herd, i.e. if you actually can sell it. Once you sold it, the button should disappear. 
If you still see the button, it might be related to a bug I discovered while building this. I described it here https://app.zenhub.com/workspaces/aspebodaklungan-60507bca37b5030019f0c5a4/issues/nbisweden/theherdbook/334 but haven't fixed it yet.

frontend/src/individual_sellingform.tsx
HenrikeW: `herdOptions` is a way of naming that has been present in the code base since I started working with it, there are `colorOptions`, `sexOptions` and maybe some more in similar use cases, so I'd rather not change it here. 

The `herdKey` is used to reset the Autocomplete from Material UI when the selling-form-component is rendered in `individual_add`. To clear the Autocomplete when the user clicks on "Ny kanin", I change the key to cause a rerender of the component. It was the only  way I managed to clear the Autocomplete and make it show the placeholder text ("Välj besättning") and not just "undefined". Found here: https://stackoverflow.com/questions/59790956/material-ui-autocomplete-clear-value Maybe there is a better way, but I tried a lot. 
I'm open to a name it differently but thought this was most appropriate. What would you suggest?

frontend/src/individual_sellingform.tsx
HenrikeW: This change makes sense, of course. Renamed it.

pontus:
I may have been unclear and don't know if we're talking about the same button here - when selling an animal, the form wasn't automatically closed for me, and the "Sälj" button was still accessible in there (and I don't see anything that I understand would unrender it).

I never checked (at least not so I remember it ) on the individual page, but that is wrapped in a `canEdit` so I believe that is fine.

But for clarity; it's the button within the popup I'm referring to.

frontend/src/individual_sellingform.tsx
pontus: Ok, good enough.

HenrikeW:
In that case I misunderstood you. But good point! It makes sense to let that button disappear. I know replaced the user message that popped up in the bottom of the page with a success box that appears within the popup.

pontus (approved):

PR #337, opened 2021-06-29
mandatory form fields

HenrikeW:
The fields number, color, sex, litter and birth_date in the individual_form are now mandatory. 
Error handling functions have been added to individual_certificate and individual_add that both use the form.

### DIFF
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -185,7 +185,7 @@ export function IndividualAdd({
         individual.litter
       )
     ) {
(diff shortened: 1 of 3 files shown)


pontus (approved):
Untested, but I believe this should work.

I'm not thrilled about how it seems these required fields end up in so many places and would prefer being able to simply send a list of mandatory fields to `IndividualForm`, but do not directly see a way that does not get messy with references to the components.

Anyway, I believe this is good enough for merging.

Issue #343, opened 2021-06-30
fix bugs in feature for recording weight and bodyfat


HenrikeW:
Right now, it's possible to add weight and bodyfat records as an admin, but not as the owner of a rabbit. However, there are some minor bugs even for admins. 

**Major bugs ( only for owners):** 
- [x] when trying to add weight or bodyfat, the response from the request to `/api/individual` is a 502. The logs show this line: `AttributeError: 'Individual' object has no attribute 'birth_date'`
- the request body contains the whole `individual` object. Could be related to that in combination with the `admin_fields` in `form_to_individual` in `data_access.py`?
- possible fix: don't send the whole `individual` object, but only the fields that have changed?
- side note: why are `name`, `sex`, `color_note` and `color` admin fields? Shouldn't an owner be able to edit them?
- [x] side note bug: the rest of the form in `individual_edit` appears as disabled in the frontend for owners. Shouldn't be like that. Look over the permissions. This is described in #295

**Minor bugs (for both admins and owners):**
- [x] when opening the popup, the empty date input fields show errors because of missing input. Shouldn't be like that. Instead they should only show errors if the user tries to _send_ an empty field
- [x] it's not obvious how to include decimals in the weight input (3.4 works, but 3,4 doesn't)
- maybe replace unit kilogramm with gramm?
- why is there a default value (3)? 
- [x] clicking the buttons "Lägg till viktmätning" / "Lägg till hullmätning" doesn't actually do that, neither does the "delete" field. Instead you need to click the "spara" button in the lower left corner which is not obvious. Make this more intuitive.
- [x] already added bodyfat records appear in english ("normal", "low", "high"). They should be Swedish like they already are in the dropwon menu
- [x] the delete-field should say "Radera" instead of "Delete"
- [x] in the `individual_view` already recorded weights appear only if there are at least two records in the database. As soon as another one is added, both appear. Fix this.

PR #344, opened 2021-07-01
Bugfix/weight bodyfat

HenrikeW:
Fixing most of the bugs described in issue #343 

- an owner can now add weight and bodyfat records. Instead of sending the whole `individual` object to the backend, only necessary fields are sent. (OBS: the rest of the form in `individual_edit.tsx` still sends the whole object)
- when opening the popup, the date input fields are not marked with "invalid date format"-errors any more
- the request is already sent when the user clicks on "Lägg till viktmätning" / "Lägg till hullmätning". He/she doesn't need to click "Spara" any more.
- already added bodyfat records as well as the delete-field appear in Swedish instead of English. 
- in `individual_view.tsx` the weight records are shown even if there is only one weight record in the system (previously, at least two were needed)

Still unfixed: 
- The weight input field only accepts a dot as a decimal separator, but it should accept a comma instead.

### DIFF
--- a/frontend/src/individual_edit.tsx
+++ b/frontend/src/individual_edit.tsx
@@ -192,6 +192,12 @@ export function IndividualEdit({ id }: { id: string | undefined }) {
       });
   }, [genebankIndividuals]);
 
+  const translateBodyfat: Map<string, string> = new Map([
+    ["low", "låg"],
+    ["normal", "normal"],
+    ["high", "hög"],
+  ]);
+
   const asIndividual = (
     number: string | undefined
   ): LimitedIndividual | null => {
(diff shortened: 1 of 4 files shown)


frontend/src/individual_edit.tsx
```diff
@@ -244,6 +250,98 @@ export function IndividualEdit({ id }: { id: string | undefined }) {
...
+          return;
+        } else {
+          userMessage("Något gick fel.", "error");
+        }
+      })
+      .then(() => {
+        setWeight(0);
+        setWeightDate(null);
+      });
+  };
```
pontus: I'm not sure it needs addressing, but for my understanding; this doesn't handle errors if the request fails, right (e.g. browser has lost internet connection)?

pontus (approved):

frontend/src/individual_edit.tsx
HenrikeW: Right, I forgot the `catch`. Added it now, I think that should solve it.

PR #346, opened 2021-08-02
Feature/create breeding

HenrikeW:
**UPDATE**
- I moved the functionality to create and update breedings/ births into a new, separate context file. 
This way it can be used both in the breeding form and in the individual_add component. 
Much less duplicated code now. 
Also, the linter is happy now :)
-----------------------------------------------------------------------------------------------------------------------------------------------------------------
- most of the layout for creating a breeding event was already there, I just changed the UI a bit to make it hopefully more user-friendly
- there was no functionality in place in the frontend to create or update breeding events, so I added this
- I added validation for the user input to not send API requests that are doomed to fail. 
- I also added a validation to check if there are about 30 days (28-32) between birth and breeding date. If you think it's unnecessary in the frontend, I can remove it.
- I created the API requests so that they would match the available endpoints (`/api/breeding`, `api/breeding/${herdId} and `api/birth`) as they look right now, even though I don't really see the point with creating breeding and birth separately. Why not put it into one endpoint?
- I also added two lines in the backend because the birth_notes and breeding_notes weren't saved. Copy-pasted the lines from another function but it seems to work. Due to my lack of Python skills, it'd be nice if you could have an extra look here :)


frontend/src/breeding_form.tsx
```diff
@@ -116,22 +129,311 @@ export function BreedingForm({
...
+    return true;
+  };
+
+  const updateBreeding = async (breedingUpdates: Breeding) => {
+    // Only run this if there is a breeding to update.
+    if (data == "new") {
+      return;
+    }
+
+    const currentHerdBreedings = await get(`/api/breeding/${herdId}`);
```
jonandernovella: This could be in a different tsx file like we have in the backend the data_access layer.

frontend/src/breeding_form.tsx
```diff
@@ -116,22 +129,311 @@ export function BreedingForm({
...
+  const createBreeding = async (breedingData: LimitedBreeding) => {
+    const breedingCreationResponse: {
+      status: string;
+      message?: string;
+      breeding_id?: number;
+    } = await post("/api/breeding", breedingData);
+
+    if (breedingCreationResponse.status == "success") {
+      return breedingCreationResponse;
+    }
```
jonandernovella: this looks like a good workaround as long as we don't get proper http codes back

frontend/src/breeding_form.tsx
jonandernovella: The reason for this is that we can reuse the code for this api calls and data type, which is used also in other components.

jonandernovella (approved):
Good work. I left a few picky comments :)

jonandernovella:
Check the linter. There seems to be some complaint!

frontend/src/breeding_context.tsx
```diff
@@ -0,0 +1,204 @@
...
+    return breedingMatch;
+  };
+
+  /**
+   * Used before updating an existing breeding with user input.
+   * It removes empty fields and adds the breeding id.
+   * This way, the right breeding is updated and no empty fields are sent to the API.
+   * Empty fields would cause error messages from the API.
+   * */
+  const modifyBreedingUpdates = (updates: Breeding, breedingMatch) => {
```
jonandernovella, line 182: Why is breeding match not typed here?

frontend/src/breeding_form.tsx
```diff
@@ -57,21 +64,30 @@ const emptyBreeding: Breeding = {
 export function BreedingForm({
   data,
   herdId,
+  handleBreedingsChanged,
+  handleActive,
 }: {
   data: Breeding | "new";
   herdId: string | undefined;
+  handleBreedingsChanged: any;
+  handleActive: any;
```
jonandernovella: Why any? They sound like a boolean.

frontend/src/breeding_form.tsx
```diff
@@ -41,13 +48,13 @@ const useStyles = makeStyles({
...
-  breed_date: "",
+  breed_date: null,
   breed_notes: "",
   father: "",
   mother: "",
-  birth_date: "",
+  birth_date: null,
   birth_notes: "",
-  litter_size: 0,
+  litter_size: null,
```
jonandernovella, line 57: What about 0?

frontend/src/main.tsx
```diff
@@ -32,7 +33,9 @@ const Main = (
       <WithDataContext>
         <WithUserContext>
           <WithMessageContext>
-            <Navigation />
+            <WithBreedingContext>
+              <Navigation />
+            </WithBreedingContext>
```
jonandernovella, line 38: Neat!

jonandernovella (commented):
Left a few thoughts. Otherwise it looks in good shape!

frontend/src/breeding_context.tsx
HenrikeW: I don't know, actually. Added it.

frontend/src/breeding_form.tsx
HenrikeW: Nope, they are functions. But not "any" anyway, so I changed it. Thanks!

frontend/src/breeding_form.tsx
HenrikeW: I want the input field empty when the page renders. It's only empty if the value is null, otherwise it shows a 0.

HenrikeW:
Thanks for having eagle eyes, Jon! Added some types where they were wrong / missing but would like to keep it null instead of 0, see above. If you're happy with it, feel free to merge.

PR #350, opened 2021-08-06
Allow generation of new(!) individual entries

jonandernovella:
With this PR I added the backend and frontend bits needed for creating new individual identifiers that are not entered manually by an user.

-  It checks how many breeding events happened on a given year in a herd to figure out the "Kull". It assumes that this is enough information to assign the next number.
- The last number of the identifier is calculated based on the litter size of the breeding event.

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -429,9 +429,7 @@ def next_individual_number(herd, birth_date, breeding_event):
         if litter_size == 0:
             next_litter += 1
 
-        ind_number = (
-            f"{herd}-{str(birth_date.year)[2:4]}{next_litter}{litter_size + 1}"
-        )
+        ind_number = f"{herd}-{str(birth_date.year)[2:4]}{next_litter}{litter_size + 1}"
 
         return ind_number
 
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -402,7 +402,7 @@ class Breeding(BaseModel):
 
     def next_individual_number(herd, birth_date, breeding_event):
         """
-        Returns the next litter number for a given year and herd.
+        Returns the number for the next individual in a litter for a given year and herd.
         """
 
         try:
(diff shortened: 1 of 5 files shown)


app/utils/database.py
```diff
@@ -400,6 +400,53 @@ class Meta:  # pylint: disable=too-few-public-methods
...
+            # last_event_id = events[0]
+            # print("Last event id is: %s " % last_event_id)
+
+            next_litter = len(events)
+
+            individuals = (
+                Individual.select(Individual).where(
+                    Individual.breeding == str(breeding_event)
+                )
+            ).execute()
```
HenrikeW: for the first individual of a litter, this won't find anything and `individuals` will be empty/undefined/null or something, right? Will line 436 still work then and give a `litter_size` of 0 or will the `len(individuals)` lead to an error?

app/utils/database.py
```diff
@@ -400,6 +400,53 @@ class Meta:  # pylint: disable=too-few-public-methods
 
         indexes = ((("mother", "father", "birth_date"), True),)
 
+    def next_individual_number(herd, birth_date, breeding_event):
+        """
+        Returns the next litter number for a given year and herd.
+        """
```
HenrikeW: the comment and the function name sound kind of contrary to me :) what about this?
```suggestion
    def next_individual_number(herd, birth_date, breeding_event):
        """
        Returns the number for the next individual in a litter for a given year and herd.
        """
```

HenrikeW (approved):
Looks good I think. Good work :+1:  Just for clarification: this doesn't _create_ new individuals yet, when a breeding is created, right? It only _allows_ to do so, or did I miss anything?
One picky comment about a comment from me. And another one about something that might be a problem, or it might not (lack of detailed python knowledge here to determine that). 
Also, some print statements and outcommented code left that you might want to remove, for the sake of tidiness ;-)

app/utils/database.py
jonandernovella: With not finding anything do you mean that the query will return 0 records? If so, yes, `litter_size ` will be 0.

HenrikeW: alright, then it's fine :+1:

PR #351, opened 2021-08-06
Tweaks on the individual form and herd view

### LINKED ISSUE #349: Individual number validation issues
- [x] Fix validation of herd numbers. Currently the messages lie about the origin of the mother and father
- [x] Revamp the herd text field: change it to dropdown. Make a new field for the individual number.

HenrikeW:
I didn't rebase on the develop branch with the new layout but it doesn't seem to be a problem for merging.

Changes in the individual form:
- you now choose the origin herd with a dropdown and it auto-fills the first part of the individual's number
- as soon as you choose the birth date, even the first 2 of the individual's last 4 digits will show up, in line with the rabbit peoples' system for creating the individual numbers: "year, litter, individual". This hopefully helps the user to not make mistakes.
- I changed the order of input fields, mostly for design reasons.
- the validation doesn't lie about the herds any more :) I hadn't thought about that not all herds have a herd.name and my if-statement was too strict with this.

Changes in the herd view:
- the list of breeding events has moved to its own tab.
- For this, I also introduced the prop `full` to the `popup` in `message_context.tsx` to prevent the popup from changing size when switching between tabs.


airenzp:
I wanted to try these changes with the new layout so I merged it with the develop branch and added my fixes

HenrikeW:
I added some more things: 
- choose mother and father first, then origin herd options are limited to two. Adjusted layout to this, even made it (sort of) responsive
- added some extra validation for the individual numbers format

frontend/src/data_context_global.tsx
```diff
@@ -121,6 +121,20 @@ export interface Breeding {
   litter_size: number | null;
 }
 
+export interface ExtendedBreeding {
+  [key: string]: any;
+  id?: number;
+  breed_date: string | null;
+  breed_notes: string;
```
pontus: Are notes mandatory?

airenzp (approved):
I have been testing this PR and I am quite satisfied with the result.

frontend/src/data_context_global.tsx
jonandernovella: The mandatory / non-mandatory fields should be reworked. At the moment it's quite messy...

frontend/src/individual_add.tsx
```diff
@@ -449,7 +466,9 @@ export function IndividualAdd({
   const resetSibling = () => {
     const numberParts: string[] = individual?.number?.split("-");
     const sibling: Individual = {
-      number: numberParts ? numberParts[0] + "-" : null,
+      number: numberParts
+        ? numberParts[0] + "-" + numberParts[1][0] + numberParts[1][1]
```
pontus: If I undertand recent discussions (https://nbisweden.slack.com/archives/C010DDHHZ4N/p1628502747044800?thread_ts=1628500448.031700&cid=C010DDHHZ4N), the this should be numberParts[0] dash the first three digits from numberParts[1]?

frontend/src/data_context_global.tsx
```diff
@@ -121,6 +121,20 @@ export interface Breeding {
   litter_size: number | null;
 }
 
+export interface ExtendedBreeding {
```
pontus: Should probably extend `Breeding`? (https://www.typescriptlang.org/docs/handbook/2/objects.html#extending-types)

pontus (approved):
Don't see any obvious issues.

There are a few things I note, but these are old sins and I'm not sure I have the full view anyway (e.g. strings for dates vs `Date` and so on).

frontend/src/data_context_global.tsx
HenrikeW: yes, I agree. But I'm pretty sure the breed_notes shouldn't be mandatory, so I'll change it.

frontend/src/individual_add.tsx
HenrikeW: good catch! Yes, I think you're right. I'll change it.

frontend/src/data_context_global.tsx
HenrikeW: Yes, indeed. ExtendedBreeding should extend breeding. I guess I had a thought when writing this and then didn't fully think it to the end :)

jonandernovella (approved):

PR #352, opened 2021-08-11
Fix/basic layout

airenzp:
This PR continues to work on Martins basic layout. It is quite a different layout so please take a look at it and provide your comments. You can also use the test server to access it (Since nobody is using it these days I borrowed it). Now the main pages are accessed from a menu located to the left. A background image is added to all the pages, as well as a footer, etc. A new css file is added, style.css, aiming to move css code to a single file. For a start most of the styling contained in navigation.tsx and static_pages.tsx has been moved here. 
There is much more styling in the project, almost everywhere, and because these changes affect so much pages I would rather prefer to work on them on subsequent PRs (if we decide to do it), to keep this PR small. Also because I am not sure if we really want to move all the css code to a single file. Many of these styles are used by  specific tables/components. Maybe we can use the two approaches. To put the css that affects the general appearance of the pages in the stylesheet and to keep the one related with specific components in its corresponding files.

### DIFF
--- a/frontend/src/style.css
+++ b/frontend/src/style.css
@@ -140,3 +140,46 @@ list-item {
     color: #222;
   }
 }
+
+table {
+  height: 100%;
+  padding: 5px;
+  overflow-y: scroll;
+}
+
+columnLabel {
+  padding-right: 30px;
+}
+
+columnSelect {
+  z-index: 15;
+}
+
+loading {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+
+functionLink {
+  color: blue;
+  text-decoration: underline;
+  cursor: pointer;
+}
+
+sorted {
+  border: 0;
+  clip: rect(0 0 0 0);
+  height: 1;
+  margin: -1;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  top: 20;
+  width: 1;
+}
+
+search {
+  float: right;
+}
--- a/frontend/src/style.css
+++ b/frontend/src/style.css
@@ -141,34 +141,39 @@ list-item {
   }
 }
 
-table {
+.table {
   height: 100%;
   padding: 5px;
   overflow-y: scroll;
 }
 
-columnLabel {
+.sortedTable {
+  padding: 5px;
+  overflow-y: scroll;
+}
+
+.columnLabel {
   padding-right: 30px;
 }
 
-columnSelect {
+.columnSelect {
   z-index: 15;
 }
 
-loading {
+.loading {
   display: flex;
   flex-direction: column;
   justify-content: center;
   align-items: center;
 }
 
-functionLink {
+.functionLink {
   color: blue;
   text-decoration: underline;
   cursor: pointer;
 }
 
-sorted {
+.sorted {
   border: 0;
   clip: rect(0 0 0 0);
   height: 1;
@@ -180,6 +185,10 @@ sorted {
   width: 1;
 }
 
-search {
+.search {
   float: right;
+}
+
+.bottomButton {
+  float: left;
 }
--- a/frontend/src/style.css
+++ b/frontend/src/style.css
@@ -101,7 +101,7 @@ list-item {
 }
 
 .euText {
-  margin: 10px 20px;
+  margin: 10px 10px;
   display: inline;
   font-size: 0.65em;
 }
@@ -122,7 +122,7 @@ list-item {
   background: white;
   display: flex;
   opacity: 0.9;
-  text-align: justify;
+  text-align: left;
 }
 
 .heading {
@@ -137,6 +137,7 @@ list-item {
   padding-right: 4px;
 }
 
+/* desktop */
 @media screen and (min-width: 660px) {
   .main {
     margin-left: 20px;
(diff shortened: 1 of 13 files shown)


pontus:
Seems to be quite a lot of reformating changes. We (mostly @HenrikeW) have generally switched to using prettier to fixup code after saving, would that be possible to do here to avoid having large batches of changes just changing between `'` and `"` (and so on)?

airenzp:
I also moved to prettier, so we are on the same track. Just maybe these files were not edited/formatted recently?

pontus:
Looks great in general. :clap: 

Re: prettier; did you have it before or did you enable it recently? If the latter, could you try saving (giving it cause to reformat) and see if that helps the diff for `navigation.tsx`?

I also wonder if we should add a (short) prefix to the class names to distinguish what page they're used in (assuming the same name is reasonable to use for multiple pages). 

As it's now I don't think that should make a difference, but at least my vision is loading some css "externally", allowing these to be customized/overriden.

HenrikeW:
The page is starting to look really good now I think. :+1: And I think it's a good idea to move global styling into one file, and to have it in plain CSS. With the rest, I'm not sure. Even though I've also found it quite confusing to have styling in basically every file, it can have some advantages to have it in its current form as JavaScript-code instead of writing it as plain CSS. I've done some reading on it and mostly it seems to be a "more dynamic" vs "less performant" discussion. I think these articles have been interesting on it:
- this is how it's currently done in the project: https://material-ui.com/styles/basics/ 
- with some of the advantages pointed out: https://towardsdev.com/use-your-makestyles-7d347f9a3e96
- a bit of background about CSS-in-JS: https://webdesign.tutsplus.com/articles/an-introduction-to-css-in-js-examples-pros-and-cons--cms-33574
- and some of the tradeoffs: https://medium.com/free-code-camp/the-tradeoffs-of-css-in-js-bee5cf926fdb

I'm not sure what's the best way to go, but I think we should discuss it before moving a lot of stuff around and potentially be surprised by the consequences. If we decide to do it, I agree with Airen on having it in smaller PRs would be convenient.

HenrikeW:
I had another look at it and like Pontus idea of adding prefixes to the class names to make them easier to find. 
Also I wonder if there was a specific reason you chose the files you chose to start the refactoring?

HenrikeW:
And a small thing I noticed: Whenever the menu opens or closes, the whole page seems to hop slightly to the right (when opening) and back left (when closing). No idea what is causing it, maybe some padding or margin that behaves not as intended?

airenzp:
Hej, about renaming the classes with a prefix, I would rather add comments to the classes mentioning the pages where they are used, since if there are multiple references, what would be the prefix? . But I agree in that  we can rename some classes with more proper names, just that I would do that in a future PR to avoid adding extra noise to this PR.

airenzp:
We can take a look at this bug you mention Henrike later, in a zoom where you can show me.

pontus:
> Hej, about renaming the classes with a prefix, I would rather add comments to the classes mentioning the pages where they are used, since if there are multiple references, what would be the prefix? . But I agree in that we can rename some classes with more proper names, just that I would do that in a future PR to avoid adding extra noise to this PR.

For clarity; the main reason I want this is to enable customization without needing to change the code.

If the same class is used throught, either unprefixed or `c_` or `common_` or similar?

It's fine with me to defer to a separate PR, though.

airenzp:
From what I have read and seen in this PR, I think we should continue to move all the css code to a single file. If any dynamic behaviour appears, where dynamic css is feasible, we can keep the dynamic css. I find these links useful to get a quick overview about it: 
https://reactjs.org/docs/faq-styling.html
https://www.npmjs.com/package/classnames#usage-with-reactjs

Static css is also faster to render.

If you agree with me on this, then we just need to decide how to continue to work here. Do we move al the css code at once or we do it in subsequent PRs?

pontus (approved):
:clap:

frontend/src/style.css
```diff
@@ -0,0 +1,339 @@
...
+  clip: rect(0 0 0 0);
+  height: 1;
+  margin: -1;
+  overflow: hidden;
+  padding: 0;
+  position: absolute;
+  top: 20;
+  width: 1;
+}
+
```
HenrikeW: There is another block above that is also "Used in filter_table.tsx and sorted_table.tsx" starting in line 197. 
I don't know if there is another system behind the order in which the classes appear here. If not, I suggest moving this part up so that `.sorted` and `.sortedTable` are after each other.

HenrikeW (approved):
I think it looks fine now, with the fact in mind that the PR's main purpose is to move the css (not to change anything about it). 
I made one comment about a minor thing, change it if you like :)
One other thing we should consider are proper naming conventions, now that we are moving everything to a single file. Usually, CSS class names are written in kebab-case, not camelCase. 
Also to make it more maintainable, it could be a good idea to follow the BEM conventions, explained here: https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/
Maybe instead of prefixes?
But I think we can do things like this in another PR so that you can finish this for now.

frontend/src/style.css
airenzp: I added comments now for each class, specifying where they are used, since I think this way is easier to edit and test them. I just found now the time to do it. This way we can, for example, remove classes that are not used, after we finish the css cleanup in subsequent PRs. 
I agree there is much more to do with css, like using name conventions, proper names, doing a cleanup of unused features, classes...Lets go for it.

PR #353, opened 2021-08-11
fixing the canEdit function

HenrikeW:
I removed the option to pass an individual's id as an argument to the `canEdit` function. Now you can either pass a herd id or a genebank id.

Of those cases where it was used with an individual's id, I removed some entirely where I found them to be unnecessary. In all other cases I pass a herd id instead (sometimes origin herd, sometimes current herd).


pontus (approved):
Probably ok. Can't trigger any bad case with a short testing session at least.

jonandernovella (approved):
Looks good. I was confused in the beginning about the "herd" property. I think we could have named it current_herd.

PR #358, opened 2021-08-16
Swedish text in design elements

HenrikeW:
It's not a big thing, but it will make the default text in the style elements coming from the Material UI library Swedish instead of English. 
For example:
- now it's "Rader per sida" instead of "Rows per page" in the table pagination
- it's "Inga alternativ" instead of "No options" in the dropdown when you search for something that doesn't exist

Because I had to add it to `navigation.tsx` the file was formatted with Prettier, but I moved those changes into a single commit.


pontus (approved):
👍

PR #359, opened 2021-08-16
make links in individual_view open popups

### LINKED ISSUE #354: links open behind the huge popup
When looking at an individual in a large popup (individual_view.tsx) and clicking on one of the linked individuals or herds there (e.g. mother, father, herds from the herd tracking), the new view opens _behind_ the popup.
This is barely visible and it feels like nothing happens when clicking on the link.

The link should probably open another popup instead. Downside to this: the user has no way of going back to the first individual in any easy way. (we don't have a back button and the browser's back button doesn't include the popups).

HenrikeW:
fixes #354 

it's not optimal, but if we use the popups as we do today, it's more intuitive this way, I think.


pontus (approved):
Untested, looks good.

jonandernovella (approved):

PR #364, opened 2021-08-16
Do execute the litter count query

jonandernovella:
Currently we were generating ids based on the length of the SQL query string...

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -415,7 +415,7 @@ def next_individual_number(herd, birth_date, breeding_event):
             .join(Herd, on=(Herd.id == Individual.origin_herd))
             .where((Herd.herd == herd))
             .distinct()
-        )
+        ).execute()
 
         next_litter = len(events)
 


pontus (approved):
:)

pontus:
Looks good, but looking at this, I'm unsure if the breeding events selector is correct:

```
        events = (
            Breeding.select(Breeding.id)
            .join(Individual, on=(Breeding.id == Individual.breeding))
            .join(Herd, on=(Herd.id == Individual.origin_herd))
            .where((Herd.herd == herd))
            .distinct()
        ).execute()
```

I'm not convinced `origin_herd` is the thing to use.

pontus:
(But this can be merged.)

jonandernovella:
> Looks good, but looking at this, I'm unsure if the breeding events selector is correct:
> 
> ```
>         events = (
>             Breeding.select(Breeding.id)
>             .join(Individual, on=(Breeding.id == Individual.breeding))
>             .join(Herd, on=(Herd.id == Individual.origin_herd))
>             .where((Herd.herd == herd))
>             .distinct()
>         ).execute()
> ```
> 
> I'm not convinced `origin_herd` is the thing to use.

Yep. In case we would have used the herd of the mother then the current herd should have been used. However in this case the origin one should be what matters since we look at the born individuals of an event.

pontus:
> Yep. In case we would have used the herd of the mother then the current herd should have been used. However in this case the origin one should be what matters since we look at the born individuals of an event.

I still don't follow this - the `origin_herd` would be the herd where an individual is born, correct? 

The way I read the code as is, if an individual is born in herd A and sold to herd B, any litters done in herd B will "count" for herd A rather than herd B?

PR #369, opened 2021-08-19
Add new prop to herd form

### LINKED ISSUE #368: Duplicate api calls from frontend
The following lines of code produce duplicate calls to the herd endpoint to fetch info about its individuals:

https://github.com/NBISweden/theherdbook/blob/2f153bf8c003b58d1ae30a6503f86525234c7656/frontend/src/herd_view.tsx#L107

jonandernovella:
Fixes #368

This will check if herd data is available, and if so, use it instead of fetching it again.

### DIFF
--- a/frontend/src/herdForm.tsx
+++ b/frontend/src/herdForm.tsx
@@ -73,10 +73,12 @@ export function HerdForm({
   id,
   view = "form",
   change = true,
+  fromHerd,
 }: {
   id: string | undefined;
   view: "form" | "info";
   change: boolean;
+  fromHerd: Herd | undefined;
 }) {
   const { genebanks, setGenebanks } = useDataContext();
   const { user } = useUserContext();
@@ -109,6 +111,10 @@ export function HerdForm({
     setHerd({ ...defaultValues });
     setPostalcode("");
     setPostalcity("");
+    if (fromHerd) {
+      setHerd(fromHerd)
+    }
+    else {
     if (id == "new" || !id) {
       setNew(true);
     } else {
@@ -143,6 +149,7 @@ export function HerdForm({
         (error) => console.error(error)
       );
     }
+  }
     setLoading(false);
   }, [id]);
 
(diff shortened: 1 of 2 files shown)


HenrikeW:
By doing this, we remove the option for a herd owner (that is not admin or manager) to edit information about his/her herd. Until now, there has been an [Edit] link that would open the herd form. 
I always thought it was a weird place for it, but removing it entirely is not an option either, even if it might fix the duplicate fetching. 
So before doing this, we would need to implement the herd form for owners at some other place in the page, I think. 
One option could be to extend the "Inställningar" tab to become something like "My profile" where you can change your password and edit information you want to provide about your herd?

HenrikeW (changes requested):
By doing this, we remove the option for a herd owner (that is not admin or manager) to edit information about his/her herd. Until now, there has been an [Edit] link that would open the herd form.
I always thought it was a weird place for it, but removing it entirely is not an option either, even if it might fix the duplicate fetching.
So before doing this, we would need to implement the herd form for owners at some other place in the page, I think.
One option could be to extend the "Inställningar" tab to become something like "My profile" where you can change your password and edit information you want to provide about your herd?

airenzp:
I agree with Henrike. This removes the edit herd function, not quite visible here. It sounds like a good idea actually to move it to the place Henrike suggest. I think Jon is trying to avoid the delay when loading a herd by removing this function, but I still see the delay when testing in my environment for the GX1. And I see the same write error we talked about. The pedigree is also heavy to read.

airenzp:
Is this PR related with the dev/efficiency one? Maybe there we keep this function?

frontend/src/herdForm.tsx
```diff
@@ -109,6 +111,10 @@ export function HerdForm({
     setHerd({ ...defaultValues });
     setPostalcode("");
     setPostalcity("");
+    if (fromHerd) {
+      setHerd(fromHerd)
+    }
+    else {
```
HenrikeW, line 117: Great! Good solution, I think :+1:

HenrikeW (approved):
Looks good now I think, nice work. 
It feels like herds load a little bit quicker than before, but I don't have numbers on that. 
GX1 is still incredibly slow, 1:21min it took for me. But I guess that's expected.

PR #370, opened 2021-08-19
Improve response times

jonandernovella:
Response times in large herds were improved by:

- Reducing the number of generations for the computation of the pedigree
- Avoiding a large amount of SQL queries to calculate the latest herd tracking entry

The speed-ups are generally notable on my machine. Please confirm you also found an improvement!

### DIFF
--- a/frontend/src/herd_view.tsx
+++ b/frontend/src/herd_view.tsx
@@ -68,13 +68,14 @@ export function HerdView({ id }: { id: string | undefined }) {
   const { user } = useUserContext();
   const { genebanks } = useDataContext();
   const [algo, set_algo] = React.useState("Martin" as "Martin" | "Dan");
+  const [pedigreeView, setPedigreeView] = React.useState(false as boolean);
+
+  const triggerPedigreeViewState = () => {
+    setPedigreeView(true);
+  };
   const pedigree = React.useMemo(
-<<<<<<< HEAD
-    () => herdPedigree(genebanks, id, 5, algo),
-=======
-    () => herdPedigree(genebanks, id, 3, algo),
->>>>>>> Reduce generations to 3
-    [genebanks, id, algo]
+    () => pedigreeView && herdPedigree(genebanks, id, 3, algo),
+    [genebanks, id, algo, pedigreeView]
   );
 
   React.useEffect(() => {
@@ -121,7 +122,11 @@ export function HerdView({ id }: { id: string | undefined }) {
           >
             <Tab label="Lista över individer" value="list" />
             <Tab label="Parningstillfällen" value="breeding" />
-            <Tab label="Släktträd för besättningen" value="pedigree" />
+            <Tab
+              label="Släktträd för besättningen"
+              value="pedigree"
+              onClick={triggerPedigreeViewState}
+            />
           </Tabs>
         </AppBar>
 
@@ -166,7 +171,7 @@ export function HerdView({ id }: { id: string | undefined }) {
               </select>
             </label>
           </div>
-          {pedigree && (
+          {pedigreeView && (
             <PedigreeNetwork
               pedigree={pedigree}
               onClick={(node: string) =>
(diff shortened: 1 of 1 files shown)


airenzp:
I tested this branch accessing the GX1, still not responding :-(. Good improvements though.

airenzp:
I would suggest that @norling take a look here. I had heavy calculations at the backend when building the pedigree, causing similar delay and errors and a workaround was to move the calculations to the frontend. You can take a look at the code in the pedigree.tsx. Maybe the get_herd can be implemented in the frontend?

jonandernovella:
> I tested this branch accessing the GX1, still not responding :-(. Good improvements though.

If you don't document your findings, this is going nowhere :(

jonandernovella:
The backend took for me 6 seconds less to process the data, while the frontend took also around 10 seconds less for computing the pedigree of 3 gens with GX1 data.

pontus:
I believe those changes make the behaviour for active going out of scope (animals should be considered active only if there's a "recent" herd tracking entry).

jonandernovella:
> I believe those changes make the behaviour for active going out of scope (animals should be considered active only if there's a "recent" herd tracking entry).

So do you suggest bringing back the code?

pontus:
And can we delay any pedigree calculation until they are needed?

(I'm also unsure ab

> > I believe those changes make the behaviour for active going out of scope (animals should be considered active only if there's a "recent" herd tracking entry).
> 
> So do you suggest bringing back the code?

It's well possible it can be done smarter. While we're pinging @norling, I think there were cache feature of peewee that weren't really documented (from what I can tell) that could possibly be of use.

But I think the functionality needs to be there - animals can be without death information and still be inactive because they lack recent herd tracking).

pontus:
(it would certainly be nice if this could be improved, but it doesn't seem to be the mayor contributing factor to the lag I see.)

pontus:
Why are we even calculating the pedigree ahead of time here? Can't we move that until it's actually needed to display (if selected)? I guess some oddness is happening with `id` making the memo use problematic (causing delays *before* load), but the best would probably be to delay it until it's needed.

(I'm not sure about the mentions about higher performance in the frontend, that sounds odd and could very likely be resolved if it's the case.)

jonandernovella:
> Why are we even calculating the pedigree ahead of time here? Can't we move that until it's actually needed to display (if selected)? I guess some oddness is happening with `id` making the memo use problematic (causing delays _before_ load), but the best would probably be to delay it until it's needed.
> 
> (I'm not sure about the mentions about higher performance in the frontend, that sounds odd and could very likely be resolved if it's the case.)

Yes. That's indeed one way to go about it since currently the computation lives under a hook that is executed when the view is rendered.

HenrikeW:
I just tried to load the herd view of GX1. 
In develop (so without these changes) it took 1 min 21 sec in my browser.
With these changes, it's down to 50 sec. 
And with manually adding the fixes you made in the other PR (remove double herd fetching), it's down to 33 sec. 
So definitely an improvement.
About the scope of "active" (Pontus comment) I don't know. I just assumed that a rabbit is considered "active" if its "active" (or "is_active") field is `true`. Wouldn't that be the most obvious?

jonandernovella:
The last commit makes the pedigree load upon click of the corresponding tab.

airenzp (approved):
Great! Looking good to me. I do not know much about the active state otherwise.

pontus:
Seems to behave much better. 
🎉

As for the activeness, obviously the important thing is consistency, but may be best done in a separate PR if there are suitable changes

I tried reverting the `is_active` handling. That changed timing from 12.76 to 12.83 seconds for loading from the API when running on herd.gotlandskaninen.se (this is what I consider negligible and likely less than random delays accounts for).

pontus (approved):
Approving, but I'd recommend reverting the backend changes and possibly revisiting that in the future. As far as I can see they don't affect the timing of API access in a way that matters.

PR #371, opened 2021-08-19
Feature/layout refactoring

airenzp:
In this PR I did some more refactoring of the css and fixed the footer so that it is not too big when using the the mobile. 
I keep it short, so that we can review and merge without much hassle.


pontus (approved):
👍

jonandernovella (approved):

PR #372, opened 2021-08-19
adjusting frontend certificate logic

HenrikeW:
Changing the UI in individual_edit and individual_form where users can manually input a certificate number.
As there now are two properties for `certificate` and `digital_certificate` in the local state object, we need to decide in which of them to store the input, so the user can now decide this. 
To me it seemed to be the most straight forward way from a user perspective, even though I'm not happy with the amounts of code it caused. Do you have other suggestions? 

Also changing the conditional rendering rules in individual_view, but that was straight forward. 

The changes in filter_table are a layout fix for a bug we noticed (and three lines formatted by prettier for whatever reason.)

### DIFF
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -42,6 +42,7 @@ export interface Individual extends LimitedIndividual {
   origin_herd?: HerdNameID;
   genebank?: string;
   certificate: string | null;
+  digital_certificate?: string | null;
   birth_date: string | null;
   mother: LimitedIndividual | null;
   father: LimitedIndividual | null;
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -389,11 +389,10 @@ export function FilterTable({
     setOrderBy(property);
   };
 
-  const createSortHandler = (property: keyof Individual) => (
-    event: React.MouseEvent<unknown>
-  ) => {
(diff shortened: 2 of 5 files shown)


pontus:
I don't think I understand the use case here.

Some things I believe are true:

* We should not generate a digital certificate if a paper certificate exists (the `issue_certificate` function seems to verify this and refuse if there is a paper certificate.
* Digital certificates will be issued by the issue function

Even though I respect that genebank managers may need to adjust things to correct bad data, I'm not sure when they would/should change digital certificate numbers.

HenrikeW:
I can see that it doesn't rally make sense that they would/should change digital certificate numbers, but so far I've always got the feedback that genebank managers should be able to change _everything_ when asking which permissions they should have.
Earlier, when there only was only one field for certificates that either could hold a paper certificate number or a digital certificate number, they were able to change that field, no matter what type of number it was. 
Then backend changed (I don't know when) and now I adjusted the frontend to those changes to make sure that genebank managers still have the same permissions they had before. If that isn't necessary, and genebank managers should _not_ be able to change _everything_, that's new information for me.

jonandernovella (approved):

airenzp:
I will wait until Henrike is back to take a look at this PR. I do not know much about this logic and will be faster to understand it talking to her.

airenzp (approved):
Looking good overall

PR #375, opened 2021-08-30
Use active property consistently

### LINKED ISSUE #362: Establish consensus with active attribute
We should decide between active or is_active and change the backend/frontend. This should fix the frontend filter for inactive individuals.

jonandernovella:
The active property was not used consistently across the API and frontend. This ensures that is_active is correctly used in all scenarios.


pontus (approved):
Unverified, looks okay.

PR #389, opened 2021-09-03
Pass filtered data for parent selection

jonandernovella:
Depending on the use case and the role of the user, we pass different data to the individual, test breeding and breeding forms. 

- The genebank manager is able to select a date from which to fetch potential parents. 
- The owner only sees the active population.
- Test breeding behaves differently for owners  as they only see rabbits of the active population.

### DIFF
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -94,7 +94,7 @@ export function BreedingForm({
     );
     setActiveMalesLimited(
       toLimitedIndividuals(
-        getIndividuals("male", is_admin, genebank, fromDate, herdId)
+        getIndividuals("male", is_admin, genebank, fromDate, undefined)
       )
     );
   }, [fromDate, genebank]);
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -9,6 +9,8 @@ import {
 } from "@material-ui/pickers";
 import DateFnsUtils from "@date-io/date-fns";
 import {
+  getIndividuals,
+  toLimitedIndividuals,
   Birth,
   Breeding,
   dateFormat,
@@ -16,14 +18,13 @@ import {
   HerdNameID,
   individualLabel,
   inputVariant,
-  LimitedHerd,
   LimitedBreeding,
   locale,
-  OptionType,
   Individual,
   LimitedIndividual,
 } from "@app/data_context_global";
 
+import { useUserContext } from "@app/user_context";
 import { Button, TextField, Typography } from "@material-ui/core";
 import { useDataContext } from "./data_context";
 import { useMessageContext } from "@app/message_context";
@@ -68,10 +69,36 @@ export function BreedingForm({
   const { userMessage } = useMessageContext();
   const [formState, setFormState] = React.useState(emptyBreeding as Breeding);
   const [showBirthForm, setShowBirthForm] = React.useState(false);
+  let defaultDate = new Date();
+  defaultDate.setFullYear(defaultDate.getFullYear() - 10);
+  const [fromDate, setFromDate] = React.useState(defaultDate as Date);
+  const [activeMalesLimited, setActiveMalesLimited] = React.useState([]);
+  const [activeFemalesLimited, setActiveFemalesLimited] = React.useState([]);
+  const { user } = useUserContext();
+  const is_admin = !!user?.is_admin;
+
+  const genebank: Genebank | undefined = React.useMemo(() => {
+    return genebanks.find((g) => g.herds.find((h) => h.herd == herdId));
+  }, [genebanks]);
+
   React.useEffect(
     () => setFormState(!data || data == "new" ? emptyBreeding : data),
     [data]
   );
+
+  React.useEffect(() => {
+    setActiveFemalesLimited(
+      toLimitedIndividuals(
+        getIndividuals("female", is_admin, genebank, fromDate, herdId)
+      )
+    );
+    setActiveMalesLimited(
+      toLimitedIndividuals(
+        getIndividuals("male", is_admin, genebank, fromDate, herdId)
+      )
+    );
+  }, [fromDate, genebank]);
+
   /**
    * Sets a single key `label` in the `herd` form to `value` (if herd isn't
    * undefined).
@@ -82,32 +109,6 @@ export function BreedingForm({
   ) => {
     formState && setFormState({ ...formState, [label]: value });
   };
-
-  const genebank: Genebank | undefined = React.useMemo(() => {
-    return genebanks.find((g) => g.herds.find((h) => h.herd == herdId));
-  }, [genebanks]);
-
-  const allFemales: OptionType[] = React.useMemo(() => {
-    if (!genebank) {
-      return [];
-    }
-    return genebank?.individuals
-      .filter((i) => i.sex == "female")
-      .map((i) => {
-        return { value: i.number, label: individualLabel(i) };
-      });
-  }, [genebank]);
-
-  const allMales: OptionType[] = React.useMemo(() => {
-    if (!genebank) {
-      return [];
-    }
-    return genebank?.individuals
-      .filter((i) => i.sex == "male")
-      .map((i) => {
-        return { value: i.number, label: individualLabel(i) };
-      });
-  }, [genebank]);
 
   const autoFillBreedDate = (dateString: string) => {
     let breedDate: Date | number = new Date(dateString);
@@ -360,6 +361,23 @@ export function BreedingForm({
           <div className="formBox">
             <KeyboardDatePicker
               autoOk
+              variant="inline"
+              inputVariant={inputVariant}
+              disabled={is_admin ? false : true}
+              disableFuture={false}
+              className="simpleField"
+              label="Äldsta födelsedatum"
+              format={dateFormat}
+              value={fromDate}
+              InputLabelProps={{
+                shrink: true,
+              }}
+              onChange={(value: Date) => {
+                fromDate && setFromDate(value);
+              }}
+            />
+            <KeyboardDatePicker
+              autoOk
               disableFuture
               error={false}
               invalidDateMessage="Datumet har fel format."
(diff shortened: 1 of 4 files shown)


jhagberg:
The date picker is good I think that it should be changeable also by Herd owners.

In add mating record it's a bit confusing maybe the date picker should be shown above the male selection. And perhaps another text or if we ad explanation of it in a mouse over or similar.

frontend/src/breeding_form.tsx
```diff
@@ -68,10 +69,49 @@ export function BreedingForm({
...
   const [formState, setFormState] = React.useState(emptyBreeding as Breeding);
   const [showBirthForm, setShowBirthForm] = React.useState(false);
+  let defaultDate = new Date();
+  defaultDate.setFullYear(defaultDate.getFullYear() - 10);
+  const [fromDate, setFromDate] = React.useState(defaultDate as Date);
+  const [activeMalesLimited, setActiveMalesLimited] = React.useState([]);
+  const [activeFemalesLimited, setActiveFemalesLimited] = React.useState([]);
+  const { user } = useUserContext();
+  const is_admin = !!(user?.is_manager || user?.is_admin);
+  const is_owner = !!(user?.is_owner && user.is_owner.length > 0);
```
HenrikeW: I don't understand these limitations. The breeding form itself should only be accessible for users who are allowed to add a breeding: admins, genebank managers and the owner of a herd. Why do we need this specific check here?
Also, `is_owner` now checks if the user is owner of _any_ herd, but not any specific one. Is that intended?

frontend/src/breeding_form.tsx
```diff
@@ -68,10 +69,49 @@ export function BreedingForm({
...
     () => setFormState(!data || data == "new" ? emptyBreeding : data),
     [data]
   );
+
+  React.useEffect(() => {
+    setActiveFemalesLimited(
+      toLimitedIndividuals(
+        getIndividuals(
+          "female",
+          is_admin || is_owner,
```
HenrikeW: So with the definition of `is_admin` and `is_owner` above, this line only returns `false` if a user is not admin, not genebank manager and does not own a herd. That seems odd to me. Do we even have these users? And why would we have a limitation for them here?

frontend/src/breeding_form.tsx
```diff
@@ -419,10 +441,29 @@ export function BreedingForm({
...
               // jscpd:ignore-end
 
-              onChange={(event: any, newValue: OptionType | null) => {
-                newValue && setFormField("father", newValue.value);
+              onChange={(event: any, newValue: LimitedIndividual | null) => {
+                newValue && setFormField("father", newValue.number);
               }}
             />
+            {is_admin ||
+              (is_owner && (
```
HenrikeW: again, I don't understand why we need this limitation.

frontend/src/breeding_form.tsx
```diff
@@ -419,10 +441,29 @@ export function BreedingForm({
...
+                <KeyboardDatePicker
+                  autoOk
+                  variant="inline"
+                  inputVariant={inputVariant}
+                  disableFuture={false}
```
HenrikeW: Why would we allow future dates? That doesn't make sense in this context, does it?

frontend/src/breeding_form.tsx
```diff
@@ -419,10 +441,29 @@ export function BreedingForm({
...
+                  className="simpleField"
+                  label="Äldsta födelsedatum på far"
```
HenrikeW: I think this label might be confusing. What about "Begränsa listan på hanar till individer födda innan" ? (if this isn't too long...)
Also, layout wise, I'd try to set this above the father-dropdown.

frontend/src/data_context_global.tsx
```diff
@@ -210,17 +210,74 @@ export function asLocale(dateString?: string) {
...
+      (herdId ? i.herd.herd == origin.herd : true)
+  );
+}
+
+/**
+ * Returns alive individuals of given sex in the genebank from a given date
+ * @param genebank the genebank data to filter active individuals from
+ * @param sex the sex of the active individuals
+ * @param fromDate the latest birth date from which individuals should be retrieved
+ */
```
HenrikeW: Ahh, really good with a proper comment. :D I should become better on that as well...

frontend/src/individual_add.tsx
```diff
@@ -122,8 +125,15 @@ export function IndividualAdd({
...
   const [litterError, setLitterError] = React.useState(false as boolean);
+  let defaultDate = new Date();
+  defaultDate.setFullYear(defaultDate.getFullYear() - 10);
+  const [fromDate, setFromDate] = React.useState(defaultDate as Date);
+  const [activeMalesLimited, setActiveMalesLimited] = React.useState([]);
+  const [activeFemalesLimited, setActiveFemalesLimited] = React.useState([]);
   const { userMessage, popup } = useMessageContext();
   const { user } = useUserContext();
+  const is_admin = !!(user?.is_manager || user?.is_admin);
+  const is_owner = !!(user?.is_owner && user.is_owner.length > 0);
```
HenrikeW: around this, see my comments on `breeding_form.tsx`, they apply here as well

frontend/src/individual_add.tsx
```diff
@@ -488,10 +499,31 @@ export function IndividualAdd({
...
+                  InputLabelProps={{
+                    shrink: true,
+                  }}
+                  onChange={(value: Date) => {
+                    fromDate && setFromDate(value);
+                  }}
+                />
+              ))}
+          </MuiPickersUtilsProvider>
           <h2>Lägg till härstamningen</h2>
```
HenrikeW, line 506: I think the picker should be below this headline

frontend/src/individual_add.tsx
```diff
@@ -488,10 +499,31 @@ export function IndividualAdd({
...
+          <MuiPickersUtilsProvider utils={DateFnsUtils}>
+            {is_admin ||
+              (is_owner && (
+                <KeyboardDatePicker
+                  autoOk
+                  variant="inline"
+                  inputVariant={inputVariant}
+                  disableFuture={false}
+                  className="simpleField"
+                  label="Äldsta födelsedatum"
```
HenrikeW: even here, I'd choose a different label, see `breeding_form.tsx`

frontend/src/testbreed_form.tsx
```diff
@@ -182,7 +210,29 @@ export function InbreedingForm() {
               );
             })}
         </div>
+        <div className={style.lineBreak}></div>
         <form className={style.form}>
+          <MuiPickersUtilsProvider utils={DateFnsUtils}>
```
HenrikeW: Layout wise, it would be nice if the picker was less wide. On a full screen it's hard to even see the calender icon in the right end of the picker at a first glance.

HenrikeW (commented):
As I missed the last couple of days I'm not quite sure about the reason why we need this functionality. I suppose, to improve performance and not load as many individuals in the dropdown options as before? In that case, it's probably a good idea.
Or is it to improve user experience and have shorter option lists in the dropdowns? In that case, I'm not quite sure if it really is an improvement as the date pickers add extra steps. I think it was already pretty easy to find the right rabbit in the dropdowns before this change.

Anyway, I commented some things I don't understand and some things that I think could be improved, mostly UX related.

frontend/src/breeding_form.tsx
jhagberg: Another option is to just hide it 10 years old alive rabbit should be enough

frontend/src/testbreed_form.tsx
jhagberg: Yes agree.  Could we make this setting hidden and have a gear icon like "advance" settings or similar

HenrikeW (approved):
Looks good! I think it's much better than before. Still not super happy with the layout in the breeding form, but not sure if this is the moment to be picky. Ask Jonas... ^^
Some mini change suggestions if you have time & energy:
- avoid hard coded `px` in the styling and use `em` instead, makes it (more) responsive. I know we still have px in other parts of the code, but it would be nice to not introduce more of it.
- My feeling is that it should be "DÖLJ", not "DÖLJA", but I'm not 100% sure about that.
- Also, "Filtrera hanar" might sound better than "Filtrera far" ?

Great that you removed unnecessary stuff!

frontend/src/breeding_form.tsx
```diff
@@ -396,18 +399,51 @@ export function BreedingForm({
...
-                newValue && setFormField("mother", newValue.value);
+              onChange={(event: any, newValue: LimitedIndividual | null) => {
+                newValue && setFormField("mother", newValue.number);
               }}
             />
+            <Button
+              color="primary"
+              onClick={() => setShowFromDateFilter(!showFromDateFilter)}
+            >
+              {showFromDateFilter == false ? "Filtrera far" : "Dölja"}
```
jhagberg:
```suggestion
              {showFromDateFilter == false ? "Filtrera hanar" : "Dölj"}
```

frontend/src/individual_add.tsx
```diff
@@ -503,9 +512,43 @@ export function IndividualAdd({
               <TextField {...params} label="Välj mor" variant="outlined" />
             )}
           />
+          <div className={style.datum}>
+            <Button
+              color="primary"
+              onClick={() => setShowFromDateFilter(!showFromDateFilter)}
+            >
+              {showFromDateFilter == false ? "Filtrera far" : "Dölja"}
```
jhagberg:
```suggestion
              {showFromDateFilter == false ? "Filtrera hanar" : "Dölj"}
```

frontend/src/testbreed_form.tsx
```diff
@@ -182,7 +220,42 @@ export function InbreedingForm() {
...
             })}
         </div>
+        <div className={style.lineBreak}></div>
         <form className={style.form}>
+          <div className={style.datum}>
+            <Button
+              color="primary"
+              onClick={() => setShowFromDateFilter(!showFromDateFilter)}
+            >
+              {showFromDateFilter == false ? "Filtrera föräldrarna" : "Dölja"}
```
jhagberg:
```suggestion
              {showFromDateFilter == false ? "Filtrera föräldrarna" : "Dölj"}
```

jhagberg (changes requested):
Good this is better. I suggested some changes. Can you fix the px issues its good in the same time.

jhagberg:
Fix "Registrera djur" form if user is manager the filter should filter both parents. 
Rename to "Filtrera kaniner" 
Move above "mor"

frontend/src/individual_add.tsx
```diff
@@ -498,26 +504,16 @@ export function IndividualAdd({
...
           <div className={style.datum}>
             <Button
               color="primary"
               onClick={() => setShowFromDateFilter(!showFromDateFilter)}
             >
-              {showFromDateFilter == false ? "Filtrera hanar" : "Dölj"}
+              {showFromDateFilter == false
+                ? is_admin
+                  ? "Filtrera kaniner"
+                  : "Filtrera honar"
```
jhagberg:
```suggestion
                  : "Filtrera hanar"
```

jhagberg (approved):

PR #391, opened 2021-09-03
Use herd id to check for permissions

jonandernovella:
Use herd id in the backend to check whether individual adding is allowed.


jhagberg:
This change will probably make other forms fail. 
Test fails at least

jhagberg (changes requested):
This change make a herd owner not able to edit any of their own animals with error "Only manager can update name"

Maybe thats another bug its the same on develop branch

jhagberg (approved):
LGTM

PR #395, opened 2021-09-06
Update Daysbetween mating and birth

jhagberg:
according to org wishes.

### DIFF
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -135,7 +135,7 @@ export function BreedingForm({
       (birthDate.getTime() - breedingDate.getTime()) / 86400000;
     if (daysBetween < 26 || daysBetween > 38) {
       userMessage(
-        "Parningsdatum och födelsedatum passar inte ihop.",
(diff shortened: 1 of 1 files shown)


jonandernovella (changes requested):
I think we should also state this requirement in the error message.

frontend/src/breeding_form.tsx
```diff
@@ -133,7 +133,7 @@ export function BreedingForm({
     const birthDate: Date | number = new Date(birthDateString);
     const daysBetween =
       (birthDate.getTime() - breedingDate.getTime()) / 86400000;
-    if (daysBetween < 28 || daysBetween > 32) {
+    if (daysBetween < 26 || daysBetween > 38) {
       userMessage(
         "Parningsdatum och födelsedatum passar inte ihop.",
```
jonandernovella:
```suggestion
        "Tiden mellan parningsdatum och födelsedatum ska vara mellan 26 och 38 dagar.",
```

jonandernovella (approved):

PR #396, opened 2021-09-06
Create empty rabbits when breeding is updated

HenrikeW:
When a breeding event is updated with birth information, empty individuals are now created if needed. 
It's needed in two cases:
- if there was no birth information at all before the update
- if the litter size is updated to a higher number (example: if it's updated from 4 to 5, create one additional empty rabbit)

### DIFF
--- a/frontend/src/breeding_form.tsx
+++ b/frontend/src/breeding_form.tsx
@@ -173,7 +173,7 @@ export function BreedingForm({
     }
 
     if (
-      (userInput.litter_size !== null || userInput.birth_notes !== "") &&
+      (userInput.litter_size !== null || userInput.birth_notes !== null) &&
       userInput.birth_date == null
     ) {
       userMessage(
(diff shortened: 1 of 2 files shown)


frontend/src/breeding_form.tsx
```diff
@@ -263,10 +265,65 @@ export function BreedingForm({
...
+          selling_date: null,
+          breeding: modifiedBreedingUpdates.id ? modifiedBreedingUpdates.id : 0,
+        };
+        for (let i = 0; i < newIndsNumber; i++) {
+          await new Promise((r) => setTimeout(r, 2000));
+          let status = createEmptyIndividual(emptyIndividual);
+          if (!status || status == "error") {
+            break;
+          }
+        }
```
jonandernovella: I think we should wrap this in a function since it's currently duplicated.

jonandernovella (changes requested):

jonandernovella (approved):

Issue #385, opened 2021-09-02
Edit Individual Hull data is listed as undefined


jhagberg:
There is hull date information but they are all undefined 
Maybe we read in the wrong data or we can't display the data correctly?

Here is how it should be interpret 

Hull
Ägarens subjektiva bedömning. n = normalhull, f = fet (överviktig), m= mager (underviktig).

(Now we have Låg Normal Hög. We could chage it to Normal, fet, mager. ) Will check this with the mangers

andkaha:
Do we want to store the Swedish words `mager`, `normal`, and `fet`?  Or do we want the English `skinny`, `normal`, and `fat`?  Or are there better words?

jhagberg:
The python code says:
```
`bodyfat` should be a list formatted like:
    [{bodyfat: 'low' | 'normal' | 'high', date: 'yyyy-mm-dd'}, [...]]
```

so n = normal f=high m=low

andkaha:
PR #397 is now ready for review.

PR #397, opened 2021-09-06
Body fat (etc.) fixes

andkaha:
This PR fixes two things:

1. Removes any empty lines in the CSV data. This gets rid of manually added empty lines in the data that would later give rise to "duplicated entries".
2. Body fat data is no longer stored as codes taken directly from the Excel data (`low`, `normal`, `high`).

Closes issue #385.

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -377,7 +377,7 @@ while [ "$year" -le 2019 ]; do
 done | psql --quiet
 
 # Fix up the body fat data
-cat <<-END_SQL
+psql --quiet<<-'END_SQL'
 	UPDATE	bodyfat
 	SET	bodyfat = 'low'
 	WHERE	bodyfat = 'm';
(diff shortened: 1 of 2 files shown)


scripts/load-gotland.sh
```diff
@@ -376,6 +376,24 @@ while [ "$year" -le 2019 ]; do
 	year=$(( year + 1 ))
 done | psql --quiet
 
+# Fix up the body fat data
+cat <<-END_SQL
```
jhagberg: For me I get a print out on the screen data is unchanged?

jhagberg (changes requested):

scripts/load-gotland.sh
jhagberg:
```suggestion
psql --quiet<<-'END_SQL'
```
Then it loads correctly

jhagberg (changes requested):
Need to run the sql also

scripts/load-gotland.sh
andkaha: Duh. Sorry.

andkaha: Yes, that's the correct thing to do.

andkaha:
Merging, as it is working as expected.

PR #398, opened 2021-09-07
moved individual css to style.css

airenzp:


jonandernovella (approved):

HenrikeW (changes requested):
These changes introduce problems with padding and margin in the individual form because the built-in styles of the Material UI elements overwrite the plain CSS styles. 
Earlier, in similar cases, we solved this with inline styles which might work here as well, even if it's bad practice. Maybe there is a better way I can't think of right now?
Anyway, this should be fixed before we merge this.

airenzp:
Made static css to work with MaterialUI by using StylesProvider injectFirst property at the main page!

frontend/src/style.css
```diff
@@ -445,3 +394,155 @@ list-item {
...
+    padding: 0 20px;
+  }
+
+  .formPane {
+    border-right: 1px solid lightgrey;
+  }
+
+  .formPane {
+    border-right: 1px solid lightgrey;
+  }
```
HenrikeW, line 525: This is a duplicate to the rows above.

frontend/src/static_pages.tsx
```diff
@@ -202,20 +202,20 @@ export function Kontakt() {
 
 export function Footer() {
   return (
-    <div className="footer">
+    <Paper className="footer">
```
HenrikeW: I wonder why you made the footer a `Paper`. Was there a specific reason? I think, now it doesn't look like a real footer any more - it doesn't stick to the bottom of the page and has space around it on the left and right. 
Not sure if we should focus on this now but I think, at least at some point, it should be improved.

frontend/src/main.tsx
```diff
@@ -9,21 +9,24 @@ import { WithDataContext } from "@app/data_context";
...
-        <WithUserContext>
-          <WithBreedingContext>
-            <WithMessageContext>
-              <Navigation />
-            </WithMessageContext>
-          </WithBreedingContext>
-        </WithUserContext>
-      </WithDataContext>
-    </BrowserRouter>
+    <StylesProvider injectFirst>
```
HenrikeW, line 16: So cool that you could solve it this way! :100:

frontend/src/individual_form.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-import { InputAdornment, makeStyles, TextField } from "@material-ui/core";
+import { InputAdornment, StylesProvider, TextField } from "@material-ui/core";
```
HenrikeW: VSCode marks the `StylesProvider` as unused in this file. Are you sure the import is needed?

HenrikeW (approved):
Some minor things that you could change/fix if you want to, but nothing severe. Looks good now, I think!

frontend/src/style.css
airenzp: Thanks, I think I might have forgotten to run the linter in my pc, that detects this duplications, weird that did not jumped on the github linter

frontend/src/static_pages.tsx
airenzp: the paper component just made round borders, basically. I moved the footer up, but maybe was better as before, I will leave it more at the bottom. The spaces where there before. We might do some redesign later.

airenzp:
@HenrikeW Now the footer is kept at the bottom of the page and looking like a footer, hopefully. I also added the user name at the top of the page after login, so the user can see a change after it gets logged in. And the resize of the window when clicking the menu is gone!

PR #399, opened 2021-09-07
remove add button from breedings for non-owners

HenrikeW:
tiny change that removes the opportunity for users to open the breedings form in herds they don't own.


airenzp (approved):

jhagberg (approved):
LGTM tested and it works for a herd owner user

Issue #401, opened 2021-09-08
Update parents


jhagberg:
As individual parents are connected to a breeding event, there needs to be special logic to update parents of an individual.

- Redirigera individ will now have the option for either creating a new breeding event or updating an existing event
- The list of possible breeding events should be based on the birth date of the edited individual
- Let the user choose whether to create a new event or update an existing one
- If the birth date of the individual is not present in the form, this should not be visible
- Only one event should be retrieved for mother and father per date
- If several rabbits tied to breeding event, do **not** delete but create new event
- If number of rabbits tied to breeding event == 1 and the correct event does not exist yet, then update the event

PR #402, opened 2021-09-08
Bug fixes: individual registration and updates

### LINKED ISSUE #393: As a herd owner in "lägg till djur" does not trigger a post
As a herd owner in "min besättning", the menu option "lägg till" doesn't work.
No POST request is triggered, and only two GETs for the mother and father. thus herd owners cant register an animal to their herds.

jonandernovella:
- Owners could not update individuals due to being limited to some admin only attributes
- The breeding context was not being used, which led to the registration of new animals to fail
- Admin attributes have now been changed to only certificates and individual numbers

### DIFF
--- a/app/utils/data_access.py
+++ b/app/utils/data_access.py
@@ -764,21 +764,13 @@ def form_to_individual(form, user=None):
         and individual.current_herd.genebank_id in user.is_manager
     )
 
-    admin_fields = [
-        "certificate",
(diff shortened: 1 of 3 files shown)


jhagberg:
Updating the parent of a rabbit does not trigger changes in breeding and so on. Bug #401

jhagberg (approved):

PR #406, opened 2021-09-08
Allow only selecting the mother's herd

### LINKED ISSUE #400: Breeding event herd listing
Breeding event should only show up in the herd that the mother belongs to at the time of birth.

jonandernovella:
This change removes the father's herd from the pool of options to select the herd of origin.


jhagberg (approved):
LGTM works. let them test the function and we will see

Issue #388, opened 2021-09-02
In register animal


jhagberg:
In register animal accesed from Owner page. If I press "öppna i eget fönster"  the windows closes and I get back to my herd page

Issue #380, opened 2021-09-02
Swedish


jhagberg:
In admin there is still some english

Like "New User"

And "validate

PR #407, opened 2021-09-08
fixed register popup link öppna i eget fonster

airenzp:
Closes issue #388 and #380

### DIFF
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -517,7 +517,7 @@ export function FilterTable({
                 variant="contained"
                 color="primary"
                 onClick={() =>
-                  popup(<IndividualAdd herdId={id} />, "/register", true)
+                  popup(<IndividualAdd herdId={id} />, undefined, true)
                 }
               >
                 Lägg till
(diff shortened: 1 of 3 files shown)


jhagberg (changes requested):
This won't work if you are only a herd owner. You are not allowed to access /register 
Can we just disable the "öppna i eget fönster" if you are a herd owner?

airenzp:
But if you are at the popup, I guess you had the rights, what should be hidden would be the lägg till. Maybe it was, I did not test this use case.

jhagberg:
Yes maybe we should enable the register rabbit to work for Herd owner also
If you ad a user with your admin rights and only give it access to a herd then you should be able to see that you cant access /register but you can access the form via the button "lägg till" on the herd page

airenzp:
Does this has any effect on the changes here? If not maybe I can merge this PR.

jhagberg:
Yes since this changes will not work for herd owner

airenzp:
Left undefined the dialogLink parameter in the popup so that the button oppna i eget fonster is hidden:

`<DialogActions>
          {dialogLink && (
            <Link to={dialogLink} target="_blank">
              <Button color="primary">Öppna i eget fönster</Button>
            </Link>
          )}
          <Button onClick={handleCloseDialog} color="primary">
            Stäng
          </Button>
        </DialogActions>
`

jhagberg (approved):

PR #414, opened 2021-09-10
Allow managers to create herds

### LINKED ISSUE #408: Genebank manager cant create new herds in their genebanks

jonandernovella:
- Added a check for managers to be able to create herds based on whether they have access to the genbank
- Tuned a bit the canEdit function so that it checks for the genbank id instead of the name
- Fixed default label shown for new herd creation


jhagberg (approved):
LGTM tested herd owner can add herd

PR #415, opened 2021-09-10
Yearly report and reporting as dead

### LINKED ISSUE #139: As a herd owner I would like to send a yearly report to my genebank manager
This should be easy nothing really to do if you do everything in the system then every record should be up to date.

Otherwise, this could generate a PDF pre-filled with the data according to the current yearly plan document: https://drive.google.com/file/d/1yzUuL3EpyeBVFOijcYzmAOi-hozXXgOD/view?usp=sharing

HenrikeW:
Users can now send in a (very basic) yearly report where they confirm that a rabbit is still active in their herd. (To prevent the rabbit from becoming inactive 365 after the latest herdtracking).
They can also mark a rabbit as dead. 
fixes #139

### DIFF
--- a/frontend/src/individual_death.tsx
+++ b/frontend/src/individual_death.tsx
@@ -1,7 +1,11 @@
 import React from "react";
 
 import { Individual, inputVariant } from "@app/data_context_global";
+import { patch } from "./communication";
+import { useMessageContext } from "@app/message_context";
 import {
+  Box,
+  Button,
   FormControl,
   FormControlLabel,
   FormLabel,
@@ -11,6 +15,7 @@ import {
   TextField,
   Typography,
 } from "@material-ui/core";
+import { CheckCircle } from "@material-ui/icons";
 import {
   MuiPickersUtilsProvider,
   KeyboardDatePicker,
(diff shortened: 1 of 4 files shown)


jhagberg (commented):
Testing with a doing yearly report 
rabbit that has the latest herd tracking date 2021-10-01 I cant select any date. 
Testing with a rabbit that has latest date 2021-09-08 I can only report on 9 and 10 of september I guess this is by design and It might be a problem becouse I have messed up my database.

frontend/src/individual_death.tsx
```diff
@@ -0,0 +1,255 @@
...
+  React.useEffect(() => {
+    if (!isDead && !deadIndividual.butchered && !deadIndividual.death_note) {
+      handleUpdateIndividual("death_date", null);
+    }
+  }, [isDead, deadIndividual.death_note, deadIndividual.butchered]);
+
+  const onSave = () => {
+    if (!isDead) {
+      userMessage(
+        "Fyll i formen först. Har kaninen inte dött, tryck 'STÄNG'",
```
jhagberg: Fyll i formen?

jhagberg (approved):
Some minor comments. 
It seems to work. This is probably unrelated but I cant create new empty individuals any longer using the mating and birth control

frontend/src/individual_death.tsx
HenrikeW: Hehe, jag var nog lite trött i fredags :D "formuläret" passar kanske bättre?

jhagberg (approved):

PR #418, opened 2021-09-13
Add minimum selling date in selling form

jonandernovella:
This will restrict the date for the new herd tracking event so that it is not before the latest one.

### DIFF
--- a/frontend/src/individual_sellingform.tsx
+++ b/frontend/src/individual_sellingform.tsx
@@ -33,6 +33,15 @@ export function IndividualSellingForm({
   onUpdateIndividual: any;
 }) {
   const style = useStyles();
+  const getMinSellingDate = () => {
+    const minDate = new Date(
+      individual.herd_tracking
+        ? individual.herd_tracking[0].date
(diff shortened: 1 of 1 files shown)


HenrikeW (changes requested):
In general this works, but for the edge cases I'd like to improve it a bit. 
Now, you can still manually enter a selling date that is before the birth date. (And as long as you haven't chosen a birth date in the top of the form, the date picker for selling date doesn't have a minDate yet). 
Best way, I think, would be to add another case in the `validateUserInput` function in `individual_add.tsx` : In case there is a selling date, make sure it's not before the birth date. 
That would prevent the user from even sending the request. 
Another, lighter alternative would be to add a `minDateMessage` to the selling date picker. The picker already turns red if the date is before the min date, but it gives no message to the user about what's wrong. To make this work, we would need to remove `helperText="Tomt om kaninen är kvar i ursprungsbesättningen"`from the date picker, otherwise the minDateMessage won't show up (silly, but I can't change it).

HenrikeW (approved):

PR #421, opened 2021-09-13
Add herd start date

### LINKED ISSUE #417: In edit herd Min besättning
Startdatum is an "invalid data format" 
Startdatum is null?

Might be import error

andkaha:
Read the herd "start date" from the auxiliary data file and insert it
into the `herd` table.  This requires us to first remove any data that
is not on the `YYYY-MM-DD` form, and then change the column format on
the `g_data2` table for the `"Start"` column, before simply doing an
`UPDATE` as for the other pieces of auxiliary data.

Resolves #417


jhagberg:
The  mellerud has start Year only in the data but that could also be loaded

jhagberg (approved):
LGTM tested and works in frontend!

PR #422, opened 2021-09-13
Allow for breeding events with a missing father

### LINKED ISSUE #260: Import Gotlandskanin create a rabbit with null parents
G1475-2011
This rabbit is missing a father but is  when listing the rabbit imported he is listed as a founder with both father and mother as null
He should be added as a regular individual since he has his father and mother specified in the xls file

When I look further it seems to affect all animals with one missing parent
all G1475-20**  

So they should not be imported as father-mother null 
I guess the correct way is to exclude them from the import altogether instead of adding them as false founders.

andkaha:
Allows for breeding events to be missing the father.

This solves the issue that e.g. `G1475-2011` was connected to a breeding event that had no parents at all.  It (and a few other animals) are now connected to breeding events with at least one parent.

The issue is arguably a data issue, as the father for the 17 affected animals were listed for each individual but itself missing.

Resolves #260

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -174,8 +174,8 @@ psql --echo-errors --quiet <<-'END_SQL'
           FROM g_data d
           GROUP BY (d."Far nr", d."Mor nr", d."Född")
       ) d
-      JOIN individual father ON d."Far nr" = father.number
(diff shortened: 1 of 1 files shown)


jhagberg:
yes its a data issue so I think the correct way not to ad them but to print out that the father is missing and report back to them to fix the data issue so we do not need to merge this?

PR #425, opened 2021-09-13
Allow parent updates of individuals

### LINKED ISSUE #401: Update parents

jonandernovella:
This was built under the assumption that users know how to find the breeding event of the individual where they made a potential typo.

- The breeding event form allows changing father and mother, doing the corresponding update to all individuals within that event
- The individual from allows the user to change the individual's breeding event with any other breeding event in the current herd


frontend/src/individual_edit.tsx
```diff
@@ -603,50 +621,36 @@ export function IndividualEdit({ id }: { id: string | undefined }) {
...
+                  getOptionLabel={(option: Breeding) =>
+                    option.father && option.mother
+                      ? "[" +
+                        option.breed_date +
+                        "] " +
+                        option.father +
+                        " - " +
+                        option.mother
+                      : breedingLabel(individual)
+                  }
```
HenrikeW: Many of the old breeding events in the data don't have a breeding date but only a birth date. In those cases, you get something like 
```
[null] G123-1234 - G567-5678
```
To prevent this from happening we could take the birth date if there is no breeding date. And specify for the user which one we are showing. 
```suggestion
                  getOptionLabel={(option: Breeding) =>
                    option.father && option.mother
                      ? "[" +
                        (option.breed_date
                          ? "Parning " + option.breed_date
                          : "Födsel " + option.birth_date) +
                        "] " +
                        option.father +
                        " - " +
                        option.mother
                      : breedingLabel(individual)
                  }
```

frontend/src/data_context_global.tsx
```diff
@@ -149,6 +149,13 @@ export function individualLabel(individual: LimitedIndividual): string {
   return label;
 }
 
+export function breedingLabel(individual: Individual): string {
+  let father = `${individual.father.number}`;
+  let mother = `${individual.mother.number}`;
+  let date = new Date(individual.birth_date)?.toISOString().split("T")[0];
+  return "[" + date + "] " + father + " - " + mother;
+}
+
```
HenrikeW: See my comment in individual_edit. 
To let this work the same way, we could do 
```suggestion
export function breedingLabel(individual: Individual): string {
  let father = `${individual.father.number}`;
  let mother = `${individual.mother.number}`;
  let date = new Date(individual.birth_date)?.toISOString().split("T")[0];
   return "[" + "Födsel " + date + "] " + father + " - " + mother;
}

```

frontend/src/breeding_form.tsx
```diff
@@ -241,6 +248,38 @@ export function BreedingForm({
...
+      modifiedBreedingUpdates
+    );
+    const updatedBreeding = await updateBreeding(modifiedBreedingUpdates);
+    if (!!updatedBreeding) {
+      if (parentsUpdate) {
+        userMessage("Parningstillfället har uppdaterats.", "success");
+      } else {
+        userMessage(
+          "Klart. Parningen redan finns i systemet så föreldrarna har inte uppdaterats!",
+          "success"
```
HenrikeW: I think proper Swedish would be 
```suggestion
          "Klart. Parningen finns redan i systemet så föreldrarna har inte uppdaterats!",
          "success"
```
:)

HenrikeW (changes requested):
Interesting solution :D 
But it works fine, with just two small things that we should change I think, see my comments.

HenrikeW (approved):

PR #428, opened 2021-09-14
Herd view table updates automatically

### LINKED ISSUE #404: Update my herd individual list
Add rabbit and update rabbit should trigger a reload of rabbit list in my herd so that the changes to rabbits and newly rabbits will show up without user have to reload the whole page.

HenrikeW:
If a user looks at a herd in herd view, the shown table in the tab "Lista över individer" will now update automatically as soon as one of the herd's individuals is updated, added or removed from the herd. 

To achieve that, the following changes were made:
- the `get_herd` function in `data_access.py` now sends more detailed information about a herd's individuals
- `herd_view.tsx` will take its information for the table of individuals from the herd instead of the big `genebanks`variable created on page load
- the `filter_table`was adjusted to be able to take in the information on `color` sent from the backend (in a different format as before). Also the table now can show digital certificate numbers.
- the tab "Lista över individer" will fetch herd data every time it's clicked
- the global `data_context` now has listeners for the herd the user is currently looking at as well as changes that are made to individuals in that herd. Changes can be made through
1. Editing an individual
2. issuing a digital certificate
3. updating a digital certificate
4. reporting an individual as dead
5. creating empty individuals when adding a breeding
6. selling an individual

All the components for these actions are now linked to the listener.

fixes #404

### DIFF
--- a/app/tests/test_data_access.py
+++ b/app/tests/test_data_access.py
@@ -176,7 +176,7 @@ class TestDataAccess(DatabaseTest):
                 if target["genebank"] not in user.accessible_genebanks:
                     target = None
                 else:
-                    target["individuals"] = [i.short_info() for i in herd.individuals]
+                    target["individuals"] = [i.as_dict() for i in herd.individuals]
 
                 if value is None or target is None:
                     self.assertEqual(value, target)
--- a/app/tests/test_database.py
+++ b/app/tests/test_database.py
@@ -308,6 +308,10 @@ class TestDatabase(DatabaseTest):
         data["mother"] = mother
         data["father"] = father
         data["color"] = self.colors[0].name
+        data["alive"] = (
+            not self.individuals[0].death_date and not self.individuals[0].death_note
+        )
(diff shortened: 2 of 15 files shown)


jhagberg (changes requested):
Cant list live active animals need to click "visa döda" and "visa inaktiva djur" to be able to list

jhagberg:
Cant no longer list GX1 herds

jonandernovella:
The GX1 loading bug seems to be on the new frontend updates. @HenrikeW tested without @jhagberg 's changes

jhagberg:
Think its because we changed short_info to as_dict and that needs to do more lookups
 and fails on big data

jhagberg:
To reproduce this
checkout develop
start your docker

Test and list GX1 
then change  line 643 in data_access to 
        `   data["individuals"] = [i.as_dict() for i in herd.individuals]` 

try list GX1 again

then you get 502

jhagberg (approved):
Works for now!

PR #429, opened 2021-09-14
Improve edit individual

### LINKED ISSUE #384: Edit individual
Can the save button only be active if you do some changes on the data.  
Maybe a "AVBRYT" button next to save also is a good thing

HenrikeW:
A patch request to update the individual will only be sent if the user actually made changes to the individual.
Also, next to the "Spara" button there is now an "Avbryt" button that just closes the dialog. 
fixes #384

### DIFF
--- a/frontend/src/individual_edit.tsx
+++ b/frontend/src/individual_edit.tsx
@@ -455,6 +455,7 @@ export function IndividualEdit({ id }: { id: string | undefined }) {
         switch (retval.status) {
           case "success":
             userMessage(retval.message ?? "Individual updated", "success");
+            handleCloseDialog();
             break;
           default:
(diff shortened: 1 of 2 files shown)


jhagberg (changes requested):
Looks like it works as it should. Can save also close the edit dialog when clicked?

frontend/src/individual_edit.tsx
```diff
@@ -842,13 +845,21 @@ export function IndividualEdit({ id }: { id: string | undefined }) {
             </div>
             <div className={style.paneControls}>
               <Button
+                className={style.button}
                 variant="contained"
                 color="primary"
```
jhagberg, line 851: Should it be color "warning" like red? instead?

jhagberg (approved):

PR #433, opened 2021-09-14
Feature/fix is active

### LINKED ISSUE #424: Load data is_active is null
We have a missmatch Jons data have is_active by default true as in the python model. When I load is_active is empty.
Should this be handeld by the python db.init() or in the loading script?

GX1 and MX1 individuals should then be set is_active to false when importing

And other individuals that not fit the rules in Def active in database.py . 
```python
is_active = (
            self.is_active
            and not self.death_date
            and not self.death_note
            and self.latest_herdtracking_entry
            and (
                self.latest_herdtracking_entry.herd_tracking_date
                > (datetime.now() - timedelta(days=366)).date()
            )
        )
```

andkaha:
This sets the `is_active` attribute for the GX1 and MX1 herds to `false`.  The commits also fixes an oversight that made the later parts of the loading scripts overwrite some herd-related info that was set earlier.

This is related to issue #424

### DIFF
--- a/scripts/load-gotland.sh
+++ b/scripts/load-gotland.sh
@@ -128,8 +128,8 @@ psql --echo-errors --quiet <<-'END_SQL'
 			FROM genebank g, color_data c;
 
 	-- Dummy herd for individuals sold outside of the genebank
-	INSERT INTO herd (genebank_id, herd, herd_name)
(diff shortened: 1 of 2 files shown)


andkaha:
This will not be merged.

PR #436, opened 2021-09-15
Get herd info from Organisation Google space

jhagberg:
This gets the memberships number (Herd number) information that is set in field employee on google workspace 

Grants access to any herd that is connected to the member number.
Sets username as a combination of Givename and memeber number

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ docker-compose build
 docker-compose up
 ```
 
-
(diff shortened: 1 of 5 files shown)


jonandernovella:
Now I have managed to test it and the log in seems to work for me

jonandernovella (approved):

PR #439, opened 2021-09-17
validation in breeding form

### LINKED ISSUE #438: Validate birth date is not null
In the breeding event form we need to validate that the date is not null. Currently users can send an event which contains faulty data.

HenrikeW:
- breeding form doesn't allow any more to send breeding events with birth information that is incomplete
- a misleading user message was removed (it said the breeding wasn't updated when in fact, it was.)

fixes #438


jhagberg (approved):
Testat verkar funka ändringar ser OK ut! bra jobbat!

PR #442, opened 2021-09-17
Fix/breedform parents

HenrikeW:
For Owners:
If you look at an existing breeding event, mother and father will show up the form even if they are not active any more. 
To achieve that, they are manually added to the respective option arrays for females / males that are used in the form's dropdowns.

For admins & managers:
If you edit a breeding you see all males and females, even inactive and dead in the dropdown.
If you add a new one, you still only see the ones currently active for the herd the breeding event belongs to.


jhagberg:
If mother and father is active they shows up two times in the list

jhagberg (approved):
WORKS!!!

PR #443, opened 2021-09-17
ordered by Födelsedatum by default

airenzp:


jhagberg:
Its not sorting correctly if you have any rows with parningsdatum =!null

frontend/src/breeding_list.tsx
```diff
@@ -123,6 +123,7 @@ export function BreedingList({ id }: { id: string | undefined }) {
             setActive(row);
           }}
           rowsPerPage={10}
+          orderBy="Födelsedatum"
```
HenrikeW: this needs to be `"birth_date"` instead of `"Födelsedatum"`.
Then it works.

HenrikeW (changes requested):
Change one word, then it works :)

HenrikeW:
It was such a small thing, so I added a commit for it

HenrikeW (approved):

jhagberg:
WORKS!

frontend/src/breeding_list.tsx
airenzp: ah, nice. I Dont know how it was working for me then.

PR #444, opened 2021-09-17
Remove birth date bug in edit individual

### LINKED ISSUE #441: Bug in Edit individual
If I am not using the date picker 
and write a number I get  or delete a number i get

Unhandled Runtime Error

RangeError: invalid date

HenrikeW:
The birth date picker in edit individual did throw errors as soon as one started changing the date input manually.
The reason for that was a function in data_context_global that uses the birth date variable with the date method `toISOstring`.
Manually editing the birth date in the form gave (temporarily) invalid date strings.

This PR makes sure the `toISOstring` method is only called on valid date strings.

fixes #441


jhagberg (approved):
works!
Nice with "väntar på"

PR #445, opened 2021-09-17
Get breeding event of created individual of the current year

### LINKED ISSUE #440: Bug in next_individual_number
Nex_individual number is not creating the litter number correctly for the herd.

The litter number should be based on the actual number of breeding event during the current year for the current herd.

So the first breeding event for all rabbits in the herd should be the first breeding event that happend in the herd and thus get number one.

The next litter number should be number two even if the first mother is sold to another herd.

jhagberg:
Next individual number will look for individuals that was born in the current herd during the birth year of the new individual we want to add and figure out the litter number.

To get this to work a user need to register the births and create individual in the right order. 
Maybe we could figure out and order the births so they get the right number since it s the order of births thats is important but that means wee need a way to update already created individuals and that could bee tricky.

This closes #440

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -412,8 +412,18 @@ def next_individual_number(herd, birth_date, breeding_event):
         events = (
             Breeding.select(Breeding.id)
             .join(Individual, on=(Breeding.id == Individual.breeding))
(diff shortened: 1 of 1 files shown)


jhagberg:
This is outdated

Issue #273, opened 2021-05-12
Cant edit animals get "unknown color" error


jhagberg:
When edit an animal I get error and nothing updates.

running the latest images from github

jhagberg:
If you edit an individual with color null like a newly added rabbit you will send color null and the backend treat that as an error

jhagberg:
we should have a choice in frontend that is "Färg ej vald"

PR #446, opened 2021-09-17
permission and layout bugs in edit form

HenrikeW:
In individual_edit:
- Managers can now edit certificates and numbers
- If a user can edit certificates and numbers, the layout is adjusted to that


jhagberg (approved):
Works
But when editing newly added animals we hit color none bug #273

PR #447, opened 2021-09-17
its ok to have color null

### LINKED ISSUE #273: Cant edit animals get "unknown color" error

jhagberg:
close #273 

Its okey to have null as color. 

Frontend should perhaps show a color that is "null" or not yet decided.


jhagberg:
add option in frontend

PR #448, opened 2021-09-17
Fix/edit individual non manager

airenzp:


jhagberg:
or this :
```
fields = form.get(admin_field, None)
            if fields and admin_field == "number":  # parents
                changed = (
                    form[admin_field]
                    != getattr(individual, admin_field)
                )
```

Maybe @jonandernovella can explain the logic . Just a check so that a user does not update the number directly using the api endpoint and raw data?

jhagberg:
more of this section will fail. For example if you are issuing a new certificate it thinks you are trying to edit the digital_certificate field and then the test will fail
.

jhagberg:
These changes are now in #450

PR #450, opened 2021-09-18
Allow manager to update individ number

jhagberg:
New check if form have individ id check if user can edit by id
Manager can now update number, if digital certificate exists then it will be moved to the new number.  

This PR includs some permissions bug fixes and also fixes some bugs around PDF certificate generation and so on

### DIFF
--- a/app/utils/data_access.py
+++ b/app/utils/data_access.py
@@ -32,6 +32,7 @@ from utils.database import Individual  # isort: skip
 from utils.database import User  # isort: skip
 from utils.database import Weight  # isort: skip
 from utils.database import next_individual_number  # isort: skip
+import utils.s3 as s3  # isort:skip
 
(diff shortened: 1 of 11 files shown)


app/utils/data_access.py
```diff
@@ -769,21 +769,21 @@ def form_to_individual(form, user=None):
...
-
     can_manage = user and (
         user.is_admin
         or user.is_manager
         and individual.current_herd.genebank_id in user.is_manager
     )
 
+    # If the form has an number - make sure that it points to the same individual as
+    # the number if not manager then you are allowed to update number.
+    if not can_manage and "number" in form and form["number"] != individual.number:
```
jhagberg: This will not work when adding new animals. So its need to bee rewritten or change logic

jonandernovella (approved):

PR #451, opened 2021-09-23
Adding the herd where the breeding took place in the breeding model

### LINKED ISSUE #426: In parningstillfällen list bug and select parent bug
The table show breeding events of the animals the herd is currently owning.
The table should list breeding that happened for any animal when they belong to this herd. So it should include breedings for a animal even if its sold to another herd if they where breed during the stay in the current herd.

### LINKED ISSUE #440: Bug in next_individual_number

jhagberg:
This will help to track in what herd the breeding took place. 

You need to do from the scripts catalog
docker-compose -f ../dc-db-load.yml stop
docker-compose -f ../dc-db-load.yml rm
docker-compose -f ../dc-db-load.yml build (perhaps with --no-cache) 

This does not work since we have a "duplicate key value violates unique constraint "breeding_mother_id_father_id_birth_date"
I am not sure where we define this constraints.

I have checked the xls file I guess we have same mother and father on the same date but different herds.


jhagberg:
In app certs update certs by

`yes '' | openssl req -x509 -nodes -days 365 -newkey rsa:4096 -keyout key.pem -out ca.pem`

app/herdbook.py
```diff
@@ -271,18 +271,57 @@ def breedings_from_date(birth_date):
...
+            (
+                da.next_individual_number(
+                    herd=form["breeding_herd"],
+                    birth_date=form["birth_date"],
+                    breeding_event=breeding_id,
+                )
+            )
+        )
+
+
```
jonandernovella, line 324: Could have been integrated with the existing endpoint, but OK!

jonandernovella (approved):

jonandernovella (approved):

PR #454, opened 2021-10-14
Fix/new colors

jhagberg:
Moved colors to csv file since they have changed them a lot. 
Also in individual edit form color is now empty if its null. For example a newly created individual


andkaha (approved):
The changes seem to work—the loading proceeds without issue.

jonandernovella (approved):

PR #461, opened 2021-10-28
Sex is not translated in Frontend

### LINKED ISSUE #455: Show "Hona" "Hane" in frontend
In 
individual_view
herd_view 
and 
genebank_view 

Show swedish Hane and Hone instead of male and female.

jhagberg:
This closes #455 
commit 2 fixes the fillter table

### DIFF
--- a/frontend/src/filter_table.tsx
+++ b/frontend/src/filter_table.tsx
@@ -267,7 +267,12 @@ export function FilterTable({
         </a>
       ),
     },
-    { field: "sex", label: "Kön" },
+    { field: "sex",
+      label: "Kön",
+      render: (rowData: any) => (
+        rowData["sex"] == "male" ? "Hane" : rowData["sex"] == "female" ? "Hona" : ""
+        ),
+    },
     {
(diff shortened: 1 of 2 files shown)


jhagberg:
There are more places that sex is still the same as what is stored in database. 
One way is to translate in the backend but many functions in the frontend are using English values.

HenrikeW:
Works! Do you want to fix the table before merging so that #455 actually can be closed? 
Otherwise remember re-opening 455 after merging or changing the description above before merging so that the issue isn't closed and disappears without being properly resolved.

jhagberg:
Filter table is fixed in my second commit.

HenrikeW:
It works in the genebank view, where you see the big table with all rabbits. If I want to look at a herd view I get an error saying `rowData.herd is undefined`. Not sure if this is expected? If yes, it's fine to merge,

jhagberg:
Yes that error sometimes popus for me also. Did not see it now.

HenrikeW (approved):

PR #462, opened 2021-10-29
Refactor/litter size

jhagberg:
Just to make it  a bit more clear use datababase name litter_size everywhere


HenrikeW:
Everything seems to work fine except registering an animal. When I try to do that, a new breeding and birth event is created but no individual. Don't know where it goes wrong or if the bug even was introduced with this PR.
If it's gonna be addressed in a later PR this one is fine to be merged. But if the bug was introduced here, it should be fixed.

jhagberg:
Works for me

PR #463, opened 2021-10-29
Feature: Antal levande i kull vid 6 veckor

jhagberg:
Store litter_size6w in database to be able to generate digital certificates with proper data.

### DIFF
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -820,11 +820,11 @@ def update_certificate(i_number):
                 pdf_bytes=signed_data.getvalue(), ind_number=ind_data["number"]
             )
             # Update breeding if litter size has changed
-            if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
-                "litter_size6w"
-            ) != form.get("litter6w"):
-                breed_data.update(litter6w=form.get("litter6w"))
-                breed_data.update(litter=form.get("litter"))
+            if breed_data.get("litter_size") != form.get(
+                "litter_size"
+            ) or breed_data.get("litter_size6w") != form.get("litter_size6w"):
+                breed_data.update(litter_size6w=form.get("litter_size6w"))
+                breed_data.update(litter_size=form.get("litter_size"))
                 da.update_breeding(breed_data, user_id)
             da.update_individual(ind_data_copy, user_id)
     except Exception as ex:  # pylint: disable=broad-except
@@ -876,11 +876,11 @@ def issue_certificate(i_number):
         # keep the ind_data object intact
         ind_data_copy = copy.copy(ind_data)
         # Update breeding if litter size has changed
-        if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
+        if breed_data.get("litter_size") != form.get("litter_size") or breed_data.get(
             "litter_size6w"
-        ) != form.get("litter6w"):
-            breed_data.update(litter6w=form.get("litter6w"))
-            breed_data.update(litter=form.get("litter"))
+        ) != form.get("litter_size6w"):
+            breed_data.update(litter_size6w=form.get("litter_size6w"))
+            breed_data.update(litter_size=form.get("litter_size"))
             da.update_breeding(breed_data, user_id)
         res = da.update_individual(ind_data, user_id)
         cert_number = res.get("digital_certificate", None)
--- a/app/herdbook.py
+++ b/app/herdbook.py
@@ -810,7 +810,7 @@ def update_certificate(i_number):
     try:
         present = check_certificate_s3(ind_number=ind_data["number"])
         if certificate_exists and present:
-
+            breed_data = db.Breeding.get(ind_data.get("breeding")).as_dict()
             ind_data.update(**form)
             ind_data_copy = copy.copy(ind_data)
             cert_data = get_certificate_data(ind_data, user_id)
@@ -819,6 +819,13 @@ def update_certificate(i_number):
             uploaded = upload_certificate(
                 pdf_bytes=signed_data.getvalue(), ind_number=ind_data["number"]
             )
+            # Update breeding if litter size has changed
+            if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
+                "litter_size6w"
+            ) != form.get("litter6w"):
+                breed_data.update(litter6w=form.get("litter6w"))
+                breed_data.update(litter=form.get("litter"))
+                da.update_breeding(breed_data, user_id)
             da.update_individual(ind_data_copy, user_id)
     except Exception as ex:  # pylint: disable=broad-except
         APP.logger.info("Unexpected error while updating certificate: " + str(ex))
@@ -864,10 +871,17 @@ def issue_certificate(i_number):
             return jsonify({"response": "Certificate already exists"}), 400
 
         form = request.json
-
+        breed_data = db.Breeding.get(ind_data.get("breeding")).as_dict()
         ind_data.update(**form, issue_digital=True)
         # keep the ind_data object intact
         ind_data_copy = copy.copy(ind_data)
+        # Update breeding if litter size has changed
+        if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
+            "litter_size6w"
+        ) != form.get("litter6w"):
+            breed_data.update(litter6w=form.get("litter6w"))
+            breed_data.update(litter=form.get("litter"))
+            da.update_breeding(breed_data, user_id)
         res = da.update_individual(ind_data, user_id)
         cert_number = res.get("digital_certificate", None)
         cert_data = get_certificate_data(ind_data_copy, user_id)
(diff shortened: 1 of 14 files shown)


frontend/src/individual_certificate.tsx
```diff
@@ -193,6 +195,10 @@ export function IndividualCertificate({
       setLitterError(true);
       error = true;
     }
+    if (!individual?.litter6w) {
```
HenrikeW: There's a mistake here, it should be 
```suggestion
    if (!individual?.litter_size6w) {
```

frontend/src/breeding_form.tsx
```diff
@@ -409,6 +411,7 @@ export function BreedingForm({
         const newBirthData: Birth = {
           date: breeding.birth_date,
           litter_size: breeding.litter_size,
+          litter_size6w: breeding.litter_size,
```
HenrikeW, line 414: General comment about this file:
This breeding_form is used both to create and update breeding events. I think it should be possible to edit the litter_size6w in this form, but so far there is no UI for it. 
I don't understand why litter_size6w only is added for new breedings, automatically being the same value as litter_size. 
If a user adds the breeding more than 6 weeks after the birth event it should be possible to manually add litter_size6w, I think. 
And at least, it should be possible to update it later on.

Also, it could be nice to add the litter size after 6 weeks even to the breeding_list table in herd_view.

frontend/src/individual_add.tsx
```diff
@@ -328,6 +338,10 @@ export function IndividualAdd({
       userMessage("Kullstorleken måste vara störren än 0.", "warning");
       return false;
     }
+    if (individual.litter_size6w < 1) {
```
HenrikeW, line 341: It good be good to make sure that litter_size6w doesn't become too big, like
```
if (individual.litter_size6w > individual.litter_size) {
userMessage("Kullstorleken efter 6 veckor får inte vara större än kullstorleken vid födseln.", "warning");
      return false;
    }
```

frontend/src/individual_form.tsx
```diff
@@ -351,25 +353,19 @@ export function IndividualForm({
...
                   </>
                 ) : formAction == FormAction.handleCertificate ? (
+                  
                   <div className="flexRow">
-                    <KeyboardDatePicker
-                      required
-                      error={birthDateError}
-                      autoOk
-                      variant="inline"
+                  <Tooltip title="Är datumet fel vänligen ändra i parningstillfället" placement="right">
```
HenrikeW: This tooltip is only visible on hover over the input field. There is no indication that there is more information, like a small "i" in a circle or similar. If possible, it would be good to add that to improve the UI.

frontend/src/individual_form.tsx
```diff
@@ -408,6 +404,32 @@ export function IndividualForm({
...
+                  value={individual.litter_size ?? ''}
+                  type="number"
+                  onChange={(event) => {
+                    onUpdateIndividual("litter_size", +event.currentTarget.value);
+                  }}
+                />
+                <TextField
+                  required
+                  error={litterError6w}
+                  label="Antal levande i kullen (efter 6 veckor)"
```
HenrikeW: This label is too long for the form when it is used for creating or updating certificates. Maybe you could shorten it to "Antal levande i kullen efter 6v"?
An alternative would be to increase the width of the popup in which the individual_cert component is rendered, but that might be a hassle...

app/herdbook.py
```diff
@@ -819,6 +819,13 @@ def update_certificate(i_number):
             uploaded = upload_certificate(
                 pdf_bytes=signed_data.getvalue(), ind_number=ind_data["number"]
             )
+            # Update breeding if litter size has changed
+            if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
+                "litter_size6w"
+            ) != form.get("litter6w"):
+                breed_data.update(litter6w=form.get("litter6w"))
+                breed_data.update(litter=form.get("litter"))
```
HenrikeW: Didn't you refactor all the `litter`in the frontend to be called `litter-size` ? Here it's still `litter` and `litter6w`, would be good to double-check this.

app/herdbook.py
```diff
@@ -864,10 +871,17 @@ def issue_certificate(i_number):
...
         ind_data.update(**form, issue_digital=True)
         # keep the ind_data object intact
         ind_data_copy = copy.copy(ind_data)
+        # Update breeding if litter size has changed
+        if breed_data.get("litter_size") != form.get("litter") or breed_data.get(
+            "litter_size6w"
+        ) != form.get("litter6w"):
+            breed_data.update(litter6w=form.get("litter6w"))
+            breed_data.update(litter=form.get("litter"))
+            da.update_breeding(breed_data, user_id)
```
HenrikeW: More `litter` here...

app/tests/test_database.py
```diff
@@ -304,6 +304,7 @@ def test_individual(self):
 
         data["birth_date"] = self.breeding[0].birth_date
         data["litter_size"] = self.breeding[0].litter_size
+        data["litter_size6w"] = self.breeding[0].litter_size
```
HenrikeW: Why does it take the data from `self.breeding[0].litter_size` and not `self.breeding[0].litter_size6w`?

app/utils/data_access.py
```diff
@@ -1431,6 +1431,7 @@ def register_birth(form, user_uuid):
         id: <breeding database id>
         date: <birth-date, as %Y-%m-%d>,
         litter_size: <total litter size (including stillborn)>
+        litter_size6w: <number of live litter 6 weeks after birth can be null>
```
HenrikeW:
```suggestion
        litter_size6w: <number of litter alive 6 weeks after birth, can be null>
```

app/utils/data_access.py
```diff
@@ -1502,6 +1504,7 @@ def update_breeding(form, user_uuid):
         birth_date?: <date, as %Y-%m-%d>,
         birth_notes?: string,
         litter_size?: <total litter size (including stillborn)>,
+        litter_size6w: <number of live litter 6 weeks after birth can be null>
```
HenrikeW:
```suggestion
        litter_size6w: <number of litter alive 6 weeks after birth, can be null>
```

app/utils/database.py
```diff
@@ -371,6 +371,9 @@ class Breeding(BaseModel):
     """
     Table for breeding and birth.
     breeding_herd is the herd the mother existed in when she gave birth
+    litter_size6w is a field to store how many litter was alive 6 week after birth
+    if all litter of a breed exists as individuals in the herd book we can calculate this
+    but probably we need to store this for users that are only adding registrerd animals.
```
HenrikeW:
```suggestion
    litter_size6w is a field to store how many litter was alive 6 weeks after birth.
    If all litter of a breed exists as individuals in the herdbook we can calculate this
    but probably we need to store this for users that are only adding registered animals.
```

HenrikeW (changes requested):
I left some comments about 1 clear mistake, some things that might be mistakes and some ideas for improvements. 
In general I think that something is missing. Here's why:
Right now, the only way to edit the litter size after 6 weeks is by issuing or updating the certificate for one of the individuals of that litter. That's not very intuitive, I think. 
It's the same way we handle `litter_size`, but `litter_size` is something you definitely know when adding a birth or an individual and it will never change. Whereas `litter_size6w` is a value that will change potentially during 6 weeks after the birth of an individual, so I think it should be easy to edit.
My suggestion is to add it as a field to the breeding form so that users can edit it there (see comment on that file). An alternative could be to add it to the form in individual_edit.

frontend/src/individual_certificate.tsx
jhagberg: yes

frontend/src/individual_form.tsx
jhagberg: How can I do that?

app/tests/test_database.py
jhagberg: It is because I set the size6w as the same as litter_size in the test database

app/herdbook.py
jhagberg: Probable i Messed up a rebase and stash

app/herdbook.py
jhagberg: fixed

app/herdbook.py
jhagberg: Probably messed up with stash and rebase

frontend/src/breeding_form.tsx
jhagberg: The idea is that this value is only interesting when issuing a new certificate or registering a new rabbit with a paper cert. But it would be good to be able to edit it. If it was input wrong. But then you can just issue a new digital certificate with the change I think it is fine for now.

frontend/src/individual_add.tsx
jhagberg: Very good suggestion

frontend/src/individual_form.tsx
jhagberg: fixed

app/tests/test_database.py
jhagberg: I changed it so its more clear.

HenrikeW (approved):
Looks good now!

PR #464, opened 2021-10-30
Allow the secondary domain to be able to login

jhagberg:
This wont affect the UI adding the hd_domain in the login input. since only one domain is supported by google


jhagberg:
Might not be needed since other services does not allow it.

PR #465, opened 2021-10-31
Fix/id missing in yearlyreport

jhagberg:
Fixing some bug individual id not in form for sell and yearly report.


HenrikeW (approved):
Seems to work!

PR #466, opened 2021-10-31
Managers of Mellerud can now add herds in its genebank

jhagberg:


frontend/src/herdForm.tsx
```diff
@@ -35,6 +35,7 @@ import {
   FieldWithPermission,
   LimitedInputType,
 } from "@app/field_with_permission";
+import { getTarget } from "vis-util";
```
jhagberg: Do not know why this is here

HenrikeW: I was also a bit confused. Sometimes VSCode is "smart" and auto-imports things that you accidentally write in the code. I've also had some of these unwanted imports causing bugs earlier.

jhagberg: I  have removed it

HenrikeW:
I approve, but I noticed two minor things that you might wanna fix before merging: 
1) The field for mobile number in the herd form is validated like an email field. You could fix this in the `contactFields` variable in `herdForm.tsx` and give the mobile phone type "tel".
2) I tried to add a herd and put as "Besättnings-ID" a random number, like "999". When I wanted to look at the herd, it was there but without any of the information I had added. I realized then, that I was supposed to choose an ID like "M999", then it works fine. Could this become an issue for users? Then maybe prefill the field with an M or G depending on the genebank?

HenrikeW (approved):
See my comment in conversation.

HenrikeW:
It has the same error in the inputAdornment now as previous files, "Object is possibly null", but there is no workaround here. Is that added in another PR? Then this looks fine. In any case, it works.

jhagberg:
Its added here https://github.com/NBISweden/theherdbook/pull/466/commits/2117129ba6144814234b885885ac0f71389187a1
You mean error in Visual Code?

PR #468, opened 2021-11-02
Fix/register rabbits mellerud

jhagberg:
Now it finds mellerud and gotlands genebank depending on what herd we input to individual add

### DIFF
--- a/frontend/src/individual_add.tsx
+++ b/frontend/src/individual_add.tsx
@@ -194,6 +194,7 @@ export function IndividualAdd({
   React.useEffect(() => {
     if (!!genebank) {
       setCurrentGenebank(genebank);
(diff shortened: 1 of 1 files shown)


HenrikeW:
Works! But I maybe found another bug... 
If you start registering an animal for Gotlandskanin, choose a mother and father and then switch genebank to Mellerudskanin in the top, the fields for mother and father are not cleared. So it's possible to create an individual with e.g. a Gotlands-mother and a Melleruds-father. 
I approve and you can decide if you want to fix or ignore.

HenrikeW (approved):
Works! But see my comment in conversation.

HenrikeW (approved):
Nice! Works.

PR #473, opened 2021-11-24
Hotfix/register animal

jhagberg:
Managers can now sell animals to GX1 and MX1. 

Searching for breedings does no longer render an error if breeding_date i None

### DIFF
--- a/app/utils/database.py
+++ b/app/utils/database.py
@@ -970,7 +970,7 @@ class User(BaseModel, UserMixin):
         `identifier` will be parsed as:
 
         `^([a-zA-Z][0-9]+-[0-9]+)$`: individual
-        `^(([G|M]X1)|[a-zA-Z][0-9]+)$`: herd
+        `^(([GM]X1)|[a-zA-Z][0-9]+)$`: herd
         and genebank otherwise.
         """
         # admins can edit anything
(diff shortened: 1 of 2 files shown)


app/utils/database.py
```diff
@@ -970,7 +970,7 @@ def can_edit(self, identifier):
         `identifier` will be parsed as:
 
         `^([a-zA-Z][0-9]+-[0-9]+)$`: individual
-        `^([a-zA-Z][0-9]+)$`: herd
+        `^(([G|M]X1)|[a-zA-Z][0-9]+)$`: herd
```
andkaha: The regular expression `[G|M]` matches one of the three characters `G`, `|`, or `M`.  You want to use `[GM]`, or possibly just `.` (dot) to match any single character before `X1`.

app/utils/database.py
```diff
@@ -1005,7 +1005,7 @@ def can_edit(self, identifier):
             except DoesNotExist:
                 pass
 
-        elif re.match("^([a-zA-Z][0-9]+)$", identifier):
+        elif re.match("^(([G|M]X1)|[a-zA-Z][0-9]+)$", identifier):
```
andkaha: Here too, the regular expression could use `.` in place of the faulty `[G|M]`.

andkaha (changes requested):
Error in regular expressions.

app/utils/database.py
jhagberg: Aha I am mixing Javascript good catch, Yes should be GM instead

jhagberg:
```suggestion
        `^(([GM]X1)|[a-zA-Z][0-9]+)$`: herd
```

app/utils/database.py
jhagberg:
```suggestion
        elif re.match("^(([GM]X1)|[a-zA-Z][0-9]+)$", identifier):
```

andkaha (approved):
Good now.

PR #475, opened 2021-11-30
Update to set Swedish locale

jhagberg:
Before building package set swedish locale


jhagberg:
This did not solve the issues with å and so on so no need to merge

PR #476, opened 2021-11-30
fix so we do not redirect to myaccount.google for the first login

jhagberg:
Before When a user is logging in with its google account and the user is not already logged in at Google 
we redirect the user to accounts.google.[com|se] 
now we check for that and redirect to root "/"


andkaha (approved):
If it works, it's ok.  I can't easily test.

PR #477, opened 2021-12-01
Fix for å - and arrow keys being disabled in edit animal input form

jhagberg:


andkaha (approved):
Probably disables all keyboard navigation? Ok.

jhagberg:
yes its OK I think

PR #482, opened 2022-02-11
Feature/genebank update logging and general rework of submodule logging

jhagberg:


app/utils/data_access.py
```diff
@@ -36,7 +36,7 @@
 
 from werkzeug.security import check_password_hash, generate_password_hash  # isort:skip
 
-logger = logging.getLogger("herdbook.da")
```
jhagberg: Think the logger for data_access is called "da" and not "db"

airenzp (approved):

PR #483, opened 2022-02-21
check if date is not isNaN before update individual

jhagberg:
fixes warnings in backend

### DIFF
--- a/frontend/src/individual_form.tsx
+++ b/frontend/src/individual_form.tsx
@@ -253,7 +253,7 @@ export function IndividualForm({
                           shrink: true,
                         }}
                         onChange={(date, value) => {
-                          value && onUpdateIndividual("birth_date", value);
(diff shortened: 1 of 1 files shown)


HenrikeW (dismissed):
Seems reasonable. It's hard to test because when removing a birth date, what is sent is not NaN but an empty string, so I don't really know how to test this. 
However, I realized that the individual is successfully updated even when sending for example `"birth_date":"2___-__-__"` , so there doesn't seem to be proper validation of the date string. But that's probably another issue.

HenrikeW:
I tried again with adding a rabbit. The behavior I see is that whenever I fill in a birth date, a `POST` request is sent to `http://localhost:8080/api/breeding/G567` which gets a response with status 405 "Method not allowed". Doesn't seem right.

jhagberg:
Talked to @HenrikeW this is working so I will merge

PR #487, opened 2022-03-23
Locale of date picker should be in Swedish

### LINKED ISSUE #486: Calender picker is in English
Calendar date picker should be in Swedish. And week start on monday

jhagberg:
Week in date picker should start with Monday and months should be in Swedish. 
This commit looks like it fixes it.


airenzp (approved):
Works as expected. Now all the date pickers have the swedish locale

PR #488, opened 2022-04-16
Fix/npm update

jhagberg:
Some updates to for npm and build


airenzp (approved):
Looking good and working

PR #510, opened 2022-05-19
reloading the data when a parent is changed

airenzp:


jhagberg (approved):
Works. It takes time to load all individuals but it is what it is.

PR #511, opened 2022-05-19
fixing python hot reload

airenzp:
The hotreload fix needed to be done in the uwsgi_reloader. To test it:
stop the containers, add or edit a log, for example:
APP.logger.info("Connecting to database!!!!"),
start the containers and check if the updated log is printed. 
The python code is now refreshed when starting the container.


jhagberg (commented):
Now it works

PR #518, opened 2022-06-01
Fix so Admin can use the manage page

jhagberg:
(cherry picked from commit 762d20be29bb2ca2a264a6cb9fae91b15cbd4a77)


HenrikeW (approved):
Makes sense.

jhagberg:
great it works but maybe it's a more elegant way to do this

PR #524, opened 2022-06-08
similar width for register form in different situations

HenrikeW:
Media queries will always be based on the width of the complete window and not take into account whether the form lives in a popup or in a bigger window. Thus, the most feasible solution to me is to wrap the form that is not in a popup into another div, that has the same max-width as the popup (1280px). Then we can omit media queries entirely.


jhagberg (commented):
The buttons looks a bit misplaced for me.

jhagberg (approved):
Looks good now!

PR #540, opened 2022-06-28
adding color note to inidividual form and view

HenrikeW:
It's now possible to edit an individual's color note through the form used in "Redigera" and "Registrera".
The color note is also displayed in the individual view.


jhagberg (approved):
LGTM testat och funkar fint

PR #541, opened 2022-06-29
Feature/show inds in breeding

jhagberg:
This will show individuals that are linked to a breeding_id
also a user can delete empty breeding events

### DIFF
--- a/frontend/src/data_context_global.tsx
+++ b/frontend/src/data_context_global.tsx
@@ -139,7 +139,7 @@ export interface Breeding {
 export interface ExtendedBreeding extends Breeding {
   mother_name: string | null | undefined;
   father_name: string | null | undefined;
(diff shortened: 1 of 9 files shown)


frontend/src/data_context_global.tsx
```diff
@@ -139,6 +139,7 @@ export interface Breeding {
 export interface ExtendedBreeding extends Breeding {
   mother_name: string | null | undefined;
   father_name: string | null | undefined;
+  individuals: Array | null | undefined;
```
jhagberg: @HenrikeW not sure if this is Needed?

HenrikeW: Yes I think it's needed, but it caused type errors in other places. Fixed now.

HenrikeW (approved):

PR #553, opened 2022-09-01
Last import before production?

jhagberg:
Is this the last changes before we go in to production?


andkaha (approved):
Looks good. I haven't test run as I don't have updated data. Merging.

PR #635, opened 2023-04-04
fix: Rewrite test data

jhagberg:
New test data that is more real-world-like. 
Tests have been updated to reflect the changes.


jhagberg:
Test will fail for now will work with other PR

